PR #11, opened 2021-11-02
Fix up parsing more complex sources..list lines

rbarry82:
For options other than `signed-by`, parse them out into a dict so we they can be set appropriate and we don't fail. There's a surprisingly long list of options here, and while most are not relevant enough to expose, we shouldn't fail, either.

Don't raise `InvalidSourceError` unless the entire file fails to parse. This didn't come up in development since the LXD 20;04 image also has all comments stripped out of `sources.list` files.


jnsgruk:
Made some minor changes to how logging is output, and also corrected a bug where blank lines were added as repos with no details into the mapping.

PR #12, opened 2021-11-04
Check whether snap is installed at all, and catalog presence

rbarry82:
Throw a `SnapError` if `snapd` isn't installed, to provide more
meaningful feedback to users.

`snapd` does not always populate the catalog in
`/var/cache/snapd/names` at install time, especially if it's
installed/started as part of `cloud-init`, in a container, or
somewhere else where networking may not be available at first
run.

The snapd API doesn't provide a convenient way to refresh this,
and restarting the service is undesirable. Instead, if a requested
package isn't in the cache, blindly try the `snap` API to see if
we can fetch info for the requested package and return a
`SnapNotFoundError` if we get a `SnapAPIError`

### DIFF
--- a/lib/charms/operator_libs_linux/v0/snap.py
+++ b/lib/charms/operator_libs_linux/v0/snap.py
@@ -505,6 +505,8 @@ class SnapCache(Mapping):
     """
 
     def __init__(self):
+        if not self.snapd_installed:
+            raise SnapError("snapd is not installed or not in /usr/bin") from None
         self._snap_client = SnapClient()
         self._snap_map = {}
         if self.snapd_installed:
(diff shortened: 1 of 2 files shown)


lib/charms/operator_libs_linux/v0/snap.py
```diff
@@ -524,17 +525,33 @@ def __iter__(self) -> Iterable["Snap"]:
 
     def __getitem__(self, snap_name: str) -> Snap:
         """Return either the installed version or latest version for a given snap."""
+        if not self.snapd_installed:
+            raise SnapError("snapd is not installed or not in /usr/bin") from None
```
jnsgruk, line 529: Would this not be better off in `__init__`?

jnsgruk:
I have confirmed that this does in fact work, though! :)

lib/charms/operator_libs_linux/v0/snap.py
petevg: Agree. We want to catch this as quickly as possible, correct?

petevg (commented):
Agree with Jon's comment, but it's a minor nitpick. I am +1 after that is addressed (either by changing or justifying.)

rbarry82:
Happy to move it. It kills the test for it (calling `super().__init__` will try to load the "real" snap client instead of MagicMock, but it definitely does make more sense there.

jnsgruk (approved):

PR #13, opened 2021-11-04
Added systemd.py

petevg:
This is mainly a port from charmhelpers, focused on routines that
allow charm authors to stop, start, and otherwise control systemd
managed services.

This code is still a draft. It needs a proper testing strategy, and
also possibly some simplifications (this library has to do less than
the original charmhelpers host.py did).


simondeziel:
Looks interesting for potential downstream consumers, so thanks for that!

`mypy` finds a bunch of inconsistencies which would be nice to address IMHO.

petevg:
@simondeziel thanks for the comment! You're correct -- some of the types were wrong. Fixed now :-)

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,148 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+"""Abstractions for stopping, starting and managing system services via systemd.
+
+This library assumes that your charm is running on a platform that uses systemd. E.g.,
+Centos 7 or later, Ubuntu Xenial (18.04) or later.
```
simondeziel, line 19: I think Xenial had support systemd in which case, the version number would need to be fixed (18.04 -> 16.04).

petevg: Whoops! You're right, Xenial's version should be 16.04. I should remember that, because I started working at Canonical right after Xenial released :-)

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,148 @@
...
+    """Pause a system service.
+
+    Stop it, and prevent it from starting again at boot.
+
+    :param service_name: the name of the service to pause
+    """
+    stopped = True
+    if service_running(service_name):
+        stopped = service_stop(service_name)
+    service("disable", service_name)
```
simondeziel, line 130: `systemctl disable --now` will take care of stopping the service if it is running. This might be simpler.

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,148 @@
...
+    Re-enable starting again at boot. Start the service.
+
+    :param service_name: the name of the service to resume
+    """
+    service("unmask", service_name)
+    service("enable", service_name)
+    started = service_running(service_name)
+
+    if not started:
+        started = service_start(service_name)
```
simondeziel, line 147: `systemctl enable --now` will start the service if it's not running so this would be simpler.

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,148 @@
...
+    "service_resume",
+    "service_running",
+    "service_start",
+    "service_stop",
+]
+
+logger = logging.getLogger(__name__)
+
+
+def service(action: str, service_name: str) -> bool:
```
simondeziel, line 61: How about adding an optional list of arguments? This would allow passing `--quiet` for `is-active` or `--now` for `enable/disable`.

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,148 @@
...
+Centos 7 or later, Ubuntu Xenial (16.04) or later.
+
+For the most part, we transparently provide an interface to a commonly used selection of
+systemd commands, with a few shortcuts baked in. For example, service_pause and
+service_resume with run the mask/unmask and enable/disable invocations.
+
+TODO: do we want to add some magic to service_start and service_restart, so that you can
+run them without checking to see if the service is running first? E.g., service_start
+either starts or restarts a service, and service_restart either restarts or starts the
+service.
```
simondeziel, line 28: Personally, I'd stick with what systemd does natively for simplicity and to avoid confusion. My 2 cents ;)

simondeziel (commented):
I like the simplification brought by dropping support for initd/upstart.

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,143 @@
...
+
+    :param action: the action to take on the service
+    :param service_name: the name of the service to perform th action on
+    """
+    cmd = ["systemctl", action, service_name]
+    if action != "is-active":
+        logger.debug("Attempting to {} '{}' with command {}.".format(action, service_name, cmd))
+    else:
+        logger.debug("Checking if '{}' is active".format(service_name))
+    return subprocess.call(cmd) == 0
```
jnsgruk, line 67: IIRC `subprocess.call` will spew it's output into the log unless you capture it with `stdout=subprocess.PIPE, stderr=subprocess.PIPE` or similar.

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def service_running(service_name: str) -> bool:
+    """Determine whether a system service is running.
+
+    :param service_name: the name of the service
```
jnsgruk, line 73: Docstrings are formatted in the wrong style -- this should be under an `Arguments` heading. https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

Applies throughout :)

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+    return service("is-active", service_name)
+
+
+def service_start(service_name: str) -> bool:
+    """Start a system service.
+
+    :param service_name: the name of the service to stop
+    """
+    return service("start", service_name)
```
jnsgruk, line 83: Feels like each of these could do `logger.info("starting service '%s'", service_name)` or relevant message. I don't think that'd be too much and it would give some good feedback to the users of this library

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -0,0 +1,143 @@
...
+This library assumes that your charm is running on a platform that uses systemd. E.g.,
+Centos 7 or later, Ubuntu Xenial (16.04) or later.
+
+For the most part, we transparently provide an interface to a commonly used selection of
+systemd commands, with a few shortcuts baked in. For example, service_pause and
+service_resume with run the mask/unmask and enable/disable invocations.
+
+Example usage:
+```python
+from systemd import *
```
jnsgruk, line 27: I wonder if just `import systemd` is better? Feels like that + some slight renaming might be more natural? `systemd.start(service)`, `systemd.stop(service)`? I could go either way though.

jnsgruk (changes requested):
Overall nice and simple! A couple of changes needed, and ideally some integration tests now that's merged :)

lib/charms/operator_libs_linux/v0/systemd.py
petevg: I went back and forth on this. systemd.stop, systemd.start etc is pretty natural. But so is service_start, service_stop, etc. The latter more closely reflects the name of the systemd commands, and also does not imply that we are starting or stopping systemd itself.

I do realize that I tend to be enthusiastically wordy in my variable names (perhaps weirdly, for someone who uses a bare text editor with no autocomplete).

I'll give it some more thought ...

lib/charms/operator_libs_linux/v0/systemd.py
petevg: I wanted to avoid pulling in a lot of code that setup and configured the arg parse lib to handle this. We could do a naive thing, where we just accept an *args that is a list of strings that you want to pass in. That feels a bit hacky, though. 

For now, I'd like to submit the library as is, and then make a separate PR for an enhancement that allows for passing in flags.

lib/charms/operator_libs_linux/v0/systemd.py
jnsgruk: Ah yes. Silly me. Thanks

lib/charms/operator_libs_linux/v0/systemd.py
jnsgruk: Yeh, I'm not super hard over either way. Was thinking aloud :)

lib/charms/operator_libs_linux/v0/systemd.py
petevg: Updated!

tests/integration/test_systemd.py
```diff
@@ -0,0 +1,39 @@
...
+    assert service_stop("cron")
+    assert not service_running("cron")
+    assert service_start("cron")
+    assert service_running("cron")
+
+
+def test_reload():
+    # Verify that we can reload services that support reload.
+    assert not service_reload("cron")
+    assert service_reload("apparmor")
```
petevg, line 39: I'm not actually super happy about this. Trying to run reload on cron should probably produce a useful exception, rather than "False".

I might refactor this thing to throw exceptions rather than booleans.

petevg: Ah. Except "is-active" for a stopped service throws an error code, and I don't want to handle that check in Python with error codes.

I think that I'm okay w/ the True/False thing + logging to see what went wrong.

jnsgruk (approved):
Looks good - is there going to be a seperate PR that adds the LIBAPI/LIBPATCH/LIBID stuff to the library before publishing?

jnsgruk:
LGTM! 🚢 it!

PR #14, opened 2021-11-05
Basic integration tests

jnsgruk:
This PR adds basic integration tests:

- Adds the integration testing environment to `tox.ini`
- Restructure tests folder into `tests/unit` and `tests/integration`
- Add a skeleton charm for testing that uses actions to trigger the various lib methods
- Updates the Github Actions config to add integration tests


rbarry82 (approved):
Super happy to see this!

petevg (commented):
I've read through this a couple of times, and I'm not sure that I'm on board with this approach. 

Each test is essentially a hard coded bit in the test charm plus a hard coded bit in the tests. 

If we want to install an apt package, for instance, we run the action in the charm that installs an apt package. But we don't have a lot of choices about which package we're installing.

If we want to add a new test, we need to add an action on the test charm, then add a call to that action in our test file. That feels redundant, and a little confusing. 

I do see why things are written this way. Essentially, the charm is all of our actual test code, and the pytest code is a wrapper that runs the tests in a way that is reasonably well intregrated with github actions.

I'm not the biggest fan, though I have to do some thinking about what our alternatives might be. :-/

jnsgruk:
Yeh, I don't disagree that this is a little clunky. I'm not a huge fan of bundling a whole charm in the tests, but I'm not sure there is another way around it.

I did consider having a "run-tests" action that triggered each of the required tests, but you get less granularity in what passes/succeeds in the pytest output that way. 

I can see an alternative approach that would at least simplify adding test cases, and that is:

- Keep the approach of having an action per test case (these serve as nice examples, anyway)
- Remove the `set_results` and assume that if the code runs without exceptions, the test has passed and the underlying command for the packaging/passwd action has succeeded
- Use pytest's `parameterized` testing to automate just running all of the actions in the charm

This would mean that we're less coupled to outputting something in an action, then checking for the exact output in the tests, and would mean that adding a test should be as simple as just adding an action... thoughts?

jnsgruk:
Added a proof of concept for the alternative approach here: https://github.com/canonical/operator-libs-linux/pull/16/files

rbarry82:
Difficult decisions...

From my POV, I definitely prefer this patch or #16 compared to #17, since the integration test with a test charm both matches the existing patterns in OF testing, plus it matches the actual "real world" use case of the libs. I don't think running in a charm necessarily adds a lot of value since we're not exposing actions directly from the libraries (all hand-crafted just for the tests), but it's a tradeoff, since running it in a charm does at least ensure that we're in a relatively limited venv without access to "system" packages like `apt_pkg`, which could save us from potential regressions later.

Really, initializing a venv and executing the tests there may be a "best of both worlds" approach.

The explicit nature of this patch edges it out for me, barely, if we're sticking with a tester charm after others weigh in. #16's `getattr()` is good for defining the observers, but it potentially leaves anyone new to OF paradigms scratching their head as to how/why the action handlers are triggered. Since they all have to be written anyway, I'm not sure that the loss in clarity is worth the saved boilerplate.

jnsgruk:
I'm leaning more toward #17 at the moment, on the grounds that it's just simpler. And the requirements freezing isn't really an issue, because `tox` is still building environments for the tests

jnsgruk:
closing for now in favour of #17

PR #15, opened 2021-11-05
Fix GPG key handling

rbarry82:
The functions cribbed from charmhelpers assume subprocess.Popen
communication, and there are better semantics in 3.5+ with run().

Convert to use those instead


jnsgruk (approved):
Thanks, tested in a charm too.

PR #16, opened 2021-11-09
Alternative integration testing approach

jnsgruk:
This PR is demonstrative of a slightly different integration testing approach. Compared to the first iteration it:

- Makes the pytest component much lighter
- Moves assertions about the success of actions into the charm itself
- Is a little cleverer about automatically subscribing the charm to every defined action, and then automatically triggering each of them using pytest.


petevg (commented):
As indicated by my +1 on the other approach, I'm not as big of a fan of this one.

There's a lot of magic, and it's not immediately obvious how to write a test.

I really like it when I can write a test by defining a function with a relevant name. I don't mind learning a DSL in the form of a yaml file (see: the old matrix testing suite). I'm not as big a fan of understanding how two modules interplay in order to run tests. 

Regardless, thank you for taking the time to write this, rewrite it, and rewrite it once more. It was useful to see various testing strategies laid out side by side.

jnsgruk:
closing for now in favour of #17

PR #17, opened 2021-11-09
Basic Integration Testing

jnsgruk:
This approach drops juju and the test charm entirely, and just runs the tests in a lxd container.

Some work could be done with cloudinit to tidy up how the container is built, but this is just a quick poc.


tests/integration/test_apt.py
```diff
@@ -0,0 +1,46 @@
...
+        apt.update()
+        apt.add_package("zsh")
+        apt.add_package(["golang-cfssl", "jq"])
+    except apt.PackageNotFoundError:
+        logger.error("A specified package not found in package cache or on system")
+    except apt.PackageError as e:
+        logger.error(f"Could not install package. Reason: {e.message}")
+
+    paths = [get_command_path(c) for c in ["zsh", "cfssl", "jq"]]
+    assert paths == ["/usr/bin/zsh", "/usr/bin/cfssl", "/usr/bin/jq"]
```
petevg, line 26: I guess that strictly speaking, we're not 100% certain that our libraries installed these apps. If any of them make it into the distro by default in the future, our tests will cease to test.

I like the package selection here -- useful, well maintained libraries that are nonetheless probably not going to make it into the default ubuntu image in the forseeable. I think that I'm good with this.

tox.ini
```diff
@@ -60,5 +62,34 @@ deps =
...
+    lxc exec {[vars]lxd_name} -- bash -c "cloud-init status -w >/dev/null 2>&1"
+    ; Copy all the files needed for integration testing
+    lxc file push -qp {toxinidir}/tox.ini {[vars]lxd_name}/root/operator-libs-linux/
+    lxc file push -qp {toxinidir}/pyproject.toml {[vars]lxd_name}/root/operator-libs-linux/
+    lxc file push -qpr {toxinidir}/tests {[vars]lxd_name}/root/operator-libs-linux/
+    lxc file push -qpr {toxinidir}/lib {[vars]lxd_name}/root/operator-libs-linux/
+    ; Run the tests
+    lxc exec {[vars]lxd_name} -- tox -c /root/operator-libs-linux/tox.ini -e integration-tests
+commands_post =
+    -lxc stop {[vars]lxd_name}
```
petevg, line 87: I like this approach a lot. We can be flexible about where and when we run the tests, without having to actually change any of the tests themself. I suppose that we are relying on a usable version of Python being in the testing container. But again, I don't think that's a super big concern. And it's not like we're using a bunch of Python version specific features.

petevg (approved):
I am +1 on this approach. It addresses the issues at hand cleanly and succinctly, without requiring test authors to do too much thinking about where their tests are going to run, and without much danger of accidentally clobbering something on a dev box.

(If someone wants to run just one of the tests, they'll have to read and understand the tox.ini file, and/or setup a container themself. I'm okay with that, though. These are integration tests, not unit tests that need to run multiple times as one writes code.)

tox.ini
jnsgruk: The usable version of Python should get dragged in automatically by cloud-init when it installs `python3-pip` :)

jnsgruk:
I'm gonna progress with this approach. I'll do some tidying up, as the commit history has become rather messy, and also make the `passwd` tests a little more independent.

tox.ini
jameinel: I also like this approach. I was a bit worried originally, but the fact that we give you a clean place to do real mutating changes is a nice way to validate everything works as intended.

PR #18, opened 2021-11-09
Push the output from the various subprocess commands into a PIPE

jnsgruk:
At present, the `apt` library floods the `debug-log` or testing log with the output of `apt-get`, `dpkg` etc. This change pushes that output into a PIPE, and relies upon the `CalledProcessError` to relay any issues to the user.


rbarry82 (approved):
LGTM. Would be great if apt didn't spew to STDERR or `check_call` automatically stifled it, but this is  a good bypass

PR #20, opened 2021-11-11
Simpler passwd module

jnsgruk:
A proposal for a much simplified `passwd` module. Mostly a port from charm-helpers, but with some updated docstrings, and a bunch of type annotations.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/passwd.py
+++ b/lib/charms/operator_libs_linux/v0/passwd.py
@@ -33,7 +33,7 @@ import grp
 import logging
 import pwd
 from subprocess import STDOUT, check_output
-from typing import List, Optional
+from typing import List, Optional, Union
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


lib/charms/operator_libs_linux/v0/passwd.py
```diff
@@ -64,604 +45,208 @@
...
+    """
+    try:
+        user_info = pwd.getpwnam(username)
+        logger.info("user '%s' already exists", username)
+        if uid:
+            user_info = pwd.getpwuid(int(uid))
+            logger.info("user with uid '%d' already exists", uid)
+    except KeyError:
+        logger.info("creating user '%s'", username)
+        cmd = ["useradd", "--shell", shell]
```
petevg, line 128: I think the following is a bit flatter/more Pythonic. (Though I understand if you'd like to avoid returns in the middle of functions.)

```suggestion
    try:
        if uid:
            user_info = pwd.getpwnam(username)
            logger.info("user '%s' already exists", username)
            return user_info
        user_info = pwd.getpwuid(int(uid))
        logger.info("user with uid '%d' already exists", uid)
        return user_info
    except KeyError:
        logger.info("creating user '%s'", username)
        
    cmd = ["useradd", "--shell", shell]
```

lib/charms/operator_libs_linux/v0/passwd.py
```diff
@@ -64,604 +45,208 @@
...
+    shell: str = "/bin/bash",
+    system_user: bool = False,
+    primary_group: str = None,
+    secondary_groups: List[str] = None,
+    uid: int = None,
+    home_dir: str = None,
+) -> str:
+    """Add a user to the system.
+
+    Will log but otherwise succeed if the user already exists.
```
petevg, line 105: I like this. We could also name this "ensure," if we wanted to be consistent across our interfaces. But I think that "add_user" has an advantage as a name due to being shorter :-)

petevg (approved):
In general, I like this approach a lot. I provides a "this Just works" interface to commands like "add_user", and requires minimal work to implement. 

I don't like how the pwd module encourages us to raise a KeyError when a user doesn't exist. I'd much rather have a pwd.get_user method that took a default of "None". But that's not your fault.

+1 (I made some suggestions, but they aren't critical.)

jnsgruk:
Merged and published to Charmhub

PR #21, opened 2021-11-12
Use python builtin modules instead of hand-parsing for passwd

rbarry82:
The `pwd` and `grp` modules are very API-stable. Use those to
parse contents instead of regular expressions

### DIFF
--- a/lib/charms/operator_libs_linux/v0/passwd.py
+++ b/lib/charms/operator_libs_linux/v0/passwd.py
@@ -45,9 +45,9 @@ except UserNotFoundError:
 ```
 """
 
+import grp
 import logging
-import os
-import re
+import pwd
 import subprocess
 from collections import UserDict
 from enum import Enum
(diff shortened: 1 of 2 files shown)


rbarry82:
Hell is GitHub actions. I thought I opened this PR way earlier, but it was CI on my fork. Grabbing a huge act image to find out why GH actions behaves differently from a fresh LXD

rbarry82:
@petevg The alternative option for cleanup.

Removes custom handling of `/etc/passwd` and `/etc/group` with filereading and regexes, and uses builtin modules.

I don't have a strong preference either way (other than a growing dislike for GH actions, and `act` doesn't seem to be a good substitute sometimes), but it keeps the UX of the library the same as apt/snap. Choice is yours for maintenance

petevg (changes requested):
After reading through this and @jnsgruk's PR carefully, I think that I am in favor of his simpler approach. I am -1 to this, in favor of the simplest possible implementation of a wrapper around pwd.

(Feel free to tell me that you told me so if it turns out that the complexity would have helped us deal with a tricky edge case :-) )

jnsgruk:
Closing for now. Thanks for this either way.

PR #22, opened 2021-11-16
Updated from_repo_line in apt to generate valid list files

MiaAltieri:
Changed from_repo_line to convert '/' to '-' in release specifications
when generating list file
Added an additional test to verify this functionality

### DIFF
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -124,7 +124,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
 
 
 VALID_SOURCE_TYPES = ("deb", "deb-src")
@@ -934,7 +934,9 @@ class DebianRepository:
             write_file: boolean to enable writing the new repo to disk
         """
         repo = RepositoryMapping._parse(repo_line, "UserInput")
-        fname = f"{DebianRepository.prefix_from_uri(repo.uri)}-{repo.release}.list"
+        fname = (
+            f"{DebianRepository.prefix_from_uri(repo.uri)}-{repo.release.replace('/', '-')}.list"
+        )
         repo.filename = fname
 
         options = repo.options if repo.options else {}
@@ -1257,7 +1259,9 @@ class RepositoryMapping(Mapping):
           repo: a `DebianRepository` object
           default_filename: an (Optional) filename if the default is not desirable
         """
-        new_filename = f"{DebianRepository.prefix_from_uri(repo.uri)}-{repo.release}.list"
+        new_filename = (
+            f"{DebianRepository.prefix_from_uri(repo.uri)}-{repo.release.replace('/', '-')}.list"
+        )
 
         fname = repo.filename or new_filename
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_apt.py
```diff
@@ -56,3 +56,23 @@ def test_install_package_external_repository():
...
+        line = "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu focal/mongodb-org/5.0 multiverse"
+        repo = apt.DebianRepository.from_repo_line(line)
+        # Import the repository's key
+        repo.import_key(key)
+        repositories.add(repo)
+
+    apt.update()
+    apt.add_package("mongodb-org")
+
+    assert get_command_path("mongodb-org") == "/usr/bin/mongodb-org"
```
jnsgruk, line 78: Something not quite right about this test -- it's failing in the CI. Should this not be `get_command_path("mongod")`? Needs looking into

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -206,13 +206,9 @@ def __eq__(self, other) -> bool:
...
-        return (
-            isinstance(other, self.__class__)
-            and (
-                self._name,
-                self._version.number,
-            )
-            == (other._name, other._version.number)
+        return isinstance(other, self.__class__) and (self._name, self._version.number) == (
+            other._name,
+            other._version.number,
```
jnsgruk, line 211: Most of the changes in this PR seem to be formatting changes -- one to be aware of in future as it makes seeing the actual changes a little harder.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -934,7 +918,9 @@ def from_repo_line(repo_line: str, write_file: Optional[bool] = True) -> "Debian
             write_file: boolean to enable writing the new repo to disk
         """
         repo = RepositoryMapping._parse(repo_line, "UserInput")
-        fname = f"{DebianRepository.prefix_from_uri(repo.uri)}-{repo.release}.list"
+        fname = (
+            f"{DebianRepository.prefix_from_uri(repo.uri)}-{repo.release.replace('/', '-')}.list"
```
jnsgruk, line 922: I think this is right -- but I think a similar change is needed at line 1241?

jnsgruk (changes requested):
Thanks @MiaAltieri!

A couple of changes please -- would be good to have a unit test in here too that tests the name of the created repo list files with an example where it's just a plain repo like `focal` vs. one like in the test case which is `focal/something/something`

lib/charms/operator_libs_linux/v0/apt.py
MiaAltieri: Ah I see, my bad, I ran black on this and it reformatted the file. Should I revert these changes?

jnsgruk: Probably best just keep this PR clean. The code should have already been formatted through `tox -e lint` (which uses `black` with a line length of 99)

jnsgruk (approved):
Nice, thanks! One final change that would be useful in this PR is to increment the LIBPATCH by 1 so it's ready to publish once merged 🙌🏻

jnsgruk:
@rbarry82 once the LIBPATCH is updated, can you merge this and publish the latest version of Charmhub please? (I'm afk today)

rbarry82:
Definitely

jnsgruk:
Merged and published, thanks!

PR #24, opened 2021-11-22
Added snap cohorts

petevg:
Addresses gh#9

This was meant to be a quick "dive in a fix it."

We haven't actually implemented refresh, however, which means that
there is more work to do.

Next steps are to add enough code to fix the failing tests, while giving
some thought to how refresh should flow through the library in general.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/snap.py
+++ b/lib/charms/operator_libs_linux/v0/snap.py
@@ -277,7 +277,7 @@ class Snap(object):
         """
         confinement = "--classic" if self._confinement == "classic" else ""
         channel = '--channel="{}"'.format(channel) if channel else ""
-        cohort = '--cohort="{cohort}"'.format(cohort) if cohort else self._cohort
+        cohort = '--cohort="{}"'.format(cohort) if cohort else self._cohort
         self._snap("install", [confinement, channel, cohort])
 
     def _refresh(
@@ -633,7 +633,7 @@ def add(
     state: Union[str, SnapState] = SnapState.Latest,
     channel: Optional[str] = "latest",
     classic: Optional[bool] = False,
-    cohort: Optional[str] = ""
+    cohort: Optional[str] = "",
 ) -> Union[Snap, List[Snap]]:
     """Add a snap to the system.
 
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v0/snap.py
```diff
@@ -260,24 +267,51 @@ def unset(self, key) -> str:
         """
         return self._snap("unset", [key])
 
-    def _install(self, channel: Optional[str] = "") -> None:
+    def _install(self, channel: Optional[str] = "", cohort: Optional[str] = "") -> None:
         """Add a snap to the system.
 
         Args:
+          confinemenet: strict or classic
```
jnsgruk, line 274: `confinemenet` => `confinement`

Also, this doesn't actually seem to be an argument to this function?

lib/charms/operator_libs_linux/v0/snap.py
```diff
@@ -260,24 +267,51 @@ def unset(self, key) -> str:
...
           channel: the channel to install from
+          cohort: optional, the key of a cohort that this snap belongs to
         """
         confinement = "--classic" if self._confinement == "classic" else ""
```
jnsgruk, line 278: Ah, related to above, it seems we're just using the `self._confinement` so I guess that argument description can just go?

lib/charms/operator_libs_linux/v0/snap.py
```diff
@@ -260,24 +267,51 @@ def unset(self, key) -> str:
...
+            args.append('--cohort="{}"'.format(cohort))
 
-    def _refresh(self, channel: Optional[str] = "") -> None:
+        self._snap("install", args)
+
+    def _refresh(
+        self,
+        channel: Optional[str] = "",
+        cohort: Optional[str] = "",
+        leave_cohort: Optional[bool] = False,
```
jnsgruk, line 293: Could these be combined such that setting `cohort=None` is enough to imply leaving the cohort?

jnsgruk (approved):
Some minor comments, but nothing worth blocking on. Thanks!

jnsgruk:
Oh one final thing -- might be worth bumping `LIBPATCH`, and it would be nice to see an example of how one might use cohorts and refreshing to a new channel in the examples in the docstring up top :)

lib/charms/operator_libs_linux/v0/snap.py
jnsgruk: Got it - thanks!

simondeziel:
It's a bit late to comment now that it's merged but since it's a minor thing, I'll do anyway. I noticed `Optional` being used for `str` and `bool` which turn them into 3 states entities which I don't think is strictly necessary and is harder to read IMHO.

Aside from that nitpick, thanks for getting the feature in!

PR #25, opened 2021-11-22
Feature/set multiple config vals

petevg:
This breaks backwards compatibility by changing the way that snap.set and snap.get work.

I think that's okay, however, as this is a new library with no known users as of yet. And maintaining backwards compatibility with keys that contain dot notation would be difficult.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/snap.py
+++ b/lib/charms/operator_libs_linux/v0/snap.py
@@ -77,11 +77,11 @@ logger = logging.getLogger(__name__)
 LIBID = "05394e5893f94f2d90feb7cbe6b633cd"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 0
+LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
 
 
 def _cache_init(func):
--- a/lib/charms/operator_libs_linux/v0/snap.py
+++ b/lib/charms/operator_libs_linux/v0/snap.py
@@ -34,7 +34,7 @@ try:
 
     if not juju.present:
         juju.ensure(snap.SnapState.Latest, channel="beta")
-        juju.set(key="value", key2="value2")
+        juju.set({"some.key": "value", "some.key2": "value2"})
 except snap.SnapError as e:
     logger.error("An exception occurred when installing charmcraft. Reason: %s", e.message)
 ```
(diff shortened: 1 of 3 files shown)


tests/integration/test_snap.py
```diff
@@ -40,3 +40,15 @@ def test_snap_remove():
 
     snap.remove("charmcraft")
     assert get_command_path("charmcraft") == ""
+
+
+def test_snap_set():
+    lxd = snap.ensure("lxd")
+
+    assert lxd.get("ceph.external") == "false"
+    assert lxd.get("criu.enable") == "false"
```
simondeziel, line 49: FYI, trying to `get` a key that was not set yet throws an error.

simondeziel (commented):
Thanks for working on this, I have not tested it but it LGTM.

tests/integration/test_snap.py
petevg: That makes sense. I think that it should probably throw an Exception in the lib as well, though we can make it a nice custom one with a sensible error message.

tests/integration/test_snap.py
```diff
@@ -40,3 +40,32 @@ def test_snap_remove():
...
+    lxd = snap.ensure("lxd")
+
+    # Verify that the correct exception gets raised in the case of an unset key.
+    key = "keythatdoesntexist01"
+    try:
+        lxd.get(key)
+    except snap.SnapError:
+        pass
+    else:
+        logger.error("Setting an unset key should result in a SnapError.")
```
simondeziel, line 67: s/Setting/Getting/ ?

lib/charms/operator_libs_linux/v0/snap.py
```diff
@@ -34,7 +34,7 @@
 
     if not juju.present:
         juju.ensure(snap.SnapState.Latest, channel="beta")
-        juju.set("key", "value")
+        juju.set(key="value", key2="value2")
```
simondeziel, line 37: I think those 2 keys should be inside `{}` to form a single arg of Dict type.

lib/charms/operator_libs_linux/v0/snap.py
```diff
@@ -50,7 +50,7 @@
 try:
     nextcloud, charmcraft = snap.add(["nextcloud", "charmcraft"])
     if nextcloud.get("mode") != "production":
-        nextcloud.set("mode", "production")
+        nextcloud.set(mode="production")
```
simondeziel, line 53: Same as above re `{}`.

simondeziel (changes requested):

jnsgruk (approved):
Nice, thanks Pen!

PR #29, opened 2021-11-30
Add support for Python 3.5

marceloneppel:
This PR enables the support for Python 3.5 by introducing the following changes:

- Change f-strings (from Python 3.6+) to str.format().
- Change the pyfakefs to the latest version that supports Python 3.5.
- Add the same Python versions from OF CI to this repo CI. 
- Sort options dict in order to have a consistent order in Python 3.5.

Python 3.5 support is important in order to create charms like the new Postgres Machine Charm and continue to support Xenial (which uses Python 3.5).

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,9 +15,16 @@ jobs:
   unit-test:
     name: Unit tests
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
       - name: Install dependencies
         run: python -m pip install tox
       - name: Run tests
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -136,12 +136,12 @@ class Error(Exception):
 
     def __repr__(self):
         """String representation of Error."""
-        return f"<{type(self).__module__}.{type(self).__name__} {self.args}>"
+        return "<{}.{} {}>".format(type(self).__module__, type(self).__name__, self.args)
 
     @property
     def name(self):
         """Return a string representation of the model plus class."""
-        return f"<{type(self).__module__}.{type(self).__name__}>"
+        return "<{}.{}>".format(type(self).__module__, type(self).__name__)
 
     @property
     def message(self):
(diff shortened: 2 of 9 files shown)


jnsgruk (approved):
Thanks @marceloneppel! One thing that would be useful is for you to bump the LIBPATCH in each of the libraries you've modified here, which will make it easier to publish them...

rbarry82:
Thanks! Bumping LIBPATCH is definitely a great idea. Sorry for whatever you went through finding the most recent version of `pyfakefs` :/

petevg (changes requested):
Hrm. I thought that the last place we had left this conversation was that we were going to keep support for Python 3.5 in OF core, but drop support for projects like these libraries.

It's not just that I'm being grumpy about dropping f strings (though I really like f strings). I'm not sure that we have the infrastructure and personnel in place to actually make a commitment to supporting 3.5. How are we going to test it? Who is going to track security vulnerabilities and patches in any dependencies the libraries pull in? Especially given that libraries are the place where we can be less picky about what we depend on?

(I suspect that i just missed a conversation due to being out sick today. But I am kind of surprised to see this change pop up, and I am -1 on it right now.)

rbarry82:
At least in the "normal" charm world, we started stepping most of our libraries back to 3.5 also, in case they get pulled into 3.5 systems.

We've more or less gone with [static analysis on push](https://github.com/canonical/grafana-operator/commit/7904e71adaceaedf1409933134156b25443297c4#diff-cca7006f0f22de4c44b7c5ddeb1f17086a578485f0c98c49e644f6df601439eaR5) as a reasonable first pass, and doing that, at least (with a separate 3.5 environment in the GH workflow which also runs mypy) caught Jinja and a couple of other smaller deps which we had to pin back.

I definitely agree that with 3.5 being EOL, security is an open question at best, and impossible at worst, but then again, Python mostly doesn't roll its own crypto, and the (for now) intersection between `charmcraft fetch-lib` and `pip` being nonexistent makes it difficult for libs to _really_ ask for libraries from PyPi. So if we're down to stdlib/Cython only, maybe less of a concern

marceloneppel:
Thanks @jnsgruk, @rbarry82 and @petevg for the reviews. I updated the `LIBPATCH` for `apt` and `snap`. This may be a good addition for other charms that need something from Python 3.5, but I agree about the concerns @petevg, especially the security. If the PR is still relevant and if it's a best practice, I can modify to use the `static analysis` commented by @rbarry82  as a separate GH step.

jnsgruk:
Thanks @marceloneppel.

@petevg I don't disagree with you, though in this case we're actually only pinning a dev dependency used during tests (which we _could_ ultimately drop completely if we re-work the tests to use `mock_open`). I think the rest of the changes are minimal enough that the added benefit of having these libs work on Xenial for now is no bad thing? 

If we were to write libs that depended on external Python packages, and particularly any thing regarding crypto, we would have to think twice. My instinct is that `operator-libs-linux` will continue to be standalone, but we might well expand into `operator-libs-*` where more complex dependencies might come into play.

petevg (approved):

PR #31, opened 2021-12-02
Correction to test for external repository 

MiaAltieri:
Corrected repo name which we are checking for in current set of repositories


jnsgruk (approved):
Thanks :)

PR #32, opened 2021-12-03
Implements systemd daemon-reload

jnsgruk:
Implements a wrapper around `systemd daemon-reload`, and some tests to go with.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/systemd.py
+++ b/lib/charms/operator_libs_linux/v0/systemd.py
@@ -39,7 +39,7 @@ success = service_reload("nginx", restart_on_failure=True)
 import logging
 import subprocess
 
-__all__ = [  # Don't export `_systemd`. (It's not the intended way of using this lib.)
+__all__ = [  # Don't export `_systemctl`. (It's not the intended way of using this lib.)
     "service_pause",
     "service_reload",
     "service_restart",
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -174,3 +175,12 @@ def service_resume(service_name: str) -> bool:
...
     return service_running(service_name)
+
+
+def daemon_reload() -> bool:
+    """Reload systemd manager configuration."""
+    proc = subprocess.Popen(["systemctl", "daemon-reload"], **_popen_kwargs())
+    for line in iter(proc.stdout.readline, ""):
+        logger.debug(line)
+    proc.wait()
+    return proc.returncode == 0
```
petevg, line 186: Would it make more sense to just make "action" an optional argument for _service, and pass this through the _service command?

Since _service is an internal implementation detail, we could rename it something more generic. Perhaps `call`, or `ctl`.

jnsgruk: Agree, this was stupid. PR to follow...

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -83,13 +82,16 @@ def _service(action: str, service_name: str, now: bool = None, quiet: bool = Non
...
+        cmd.append(service_name)
+
     if now is not None:
         cmd.append("--now")
     if quiet is not None:
         cmd.append("--quiet")
-    if action != "is-active":
-        logger.debug("Attempting to {} '{}' with command {}.".format(action, service_name, cmd))
+    if cmd != "is-active":
+        logger.debug("Attempting to {} '{}' with command {}.".format(cmd, service_name, cmd))
```
petevg, line 94: This check will fail now, because it is comparing against a lit, rather than a string. `if 'is-active' not in cmd` would work.

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -24,23 +24,22 @@
...
-
 ```
 
 """
 
 import logging
 import subprocess
 
-__all__ = [  # Don't export `service`. (It's not the intended way of using this lib.)
+__all__ = [  # Don't export `_systemd`. (It's not the intended way of using this lib.)
```
petevg, line 42: Typo. Should be "_systemctl"

petevg (changes requested):
One broken check that I caught, and some minor typos.

petevg (approved):
LGTM!

PR #35, opened 2021-12-13
Add check for 'all' architecture

marceloneppel:
This PR adds an additional check in the `apt` library when looking for an installed package (via `dpkg`) and in the `apt-cache`. It avoids returning a `PackageNotFoundError` when there is a package with `all` architecture, like the `postgres` package (in contrast to `postgresql-12` package and others, which are listed with specific architectures like, for example, `amd64`).


petevg (approved):
Thank you for the catch, and the patch!

I am +1 to this one.

PR #36, opened 2021-12-16
Systemd library now raises Exceptions in place of returning bools

petevg:
Per the request in GH#33, we refactored the systemd lib to raise
Exceptions when it fails to do something, rather than returning false.

A path in the code keeps the behavior of is_running consistent. On a
return code of 3, it will return False for "is not running", rather
than raising an Exception.


lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -56,11 +56,15 @@
 LIBID = "045b0d179f6b4514a8bb9b48aee9ebaf"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 0
+LIBAPI = 1
```
jnsgruk, line 59: With this change, the library also needs to be put in the `lib/charms/operrator_libs_linux/v1` directory

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -98,11 +102,29 @@ def _systemctl(
...
+        if proc.returncode < 1:
+            return True
+        if proc.returncode == 3:  # Code returned when service is not active.
+            return False
+
+    if proc.returncode < 1:
+        return True
+
+    raise SystemdError(
+        "Could not {}{}. SystemD says: {}".format(
```
jnsgruk, line 124:
```suggestion
        "Could not {}{}: systemd output: {}".format(
```

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -163,9 +188,19 @@ def service_pause(service_name: str) -> bool:
...
+    except SystemdError:
+        pass
+    try:
+        _systemctl("mask", service_name)
+    except SystemdError:
+        pass
+    if not service_running(service_name):
+        return True
+    raise SystemdError(
+        "Attempted to pause {}, but it still appears to be running.".format(service_name)
```
jnsgruk, line 202:
```suggestion
        "Attempted to pause '{}', but it is still running.".format(service_name)
```

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -176,9 +211,19 @@ def service_resume(service_name: str) -> bool:
...
+        pass
+    try:
+        _systemctl("enable", service_name, now=True)
+    except SystemdError:
+        pass
+
+    if service_running(service_name):
+        return True
+
+    raise SystemdError("Attempted to resume {}, but it is not running.".format(service_name))
```
jnsgruk, line 226:
```suggestion
    raise SystemdError("Attempted to resume '{}', but it is not running.".format(service_name))
```

lib/charms/operator_libs_linux/v0/systemd.py
```diff
@@ -163,9 +188,19 @@ def service_pause(service_name: str) -> bool:
     Args:
         service_name: the name of the service to pause
     """
-    _systemctl("disable", service_name, now=True)
-    _systemctl("mask", service_name)
-    return not service_running(service_name)
+    try:
+        _systemctl("disable", service_name, now=True)
+    except SystemdError:
+        pass
```
jnsgruk, line 194: What's the rationale for this pass? (and the next for that matter?)

jnsgruk (changes requested):
Library needs moving in the `v1` directory + some minor nits on log string format

lib/charms/operator_libs_linux/v0/systemd.py
petevg: This is in keeping with the previous behavior, which was to fail silently if the mask or disable operation failed (possibly because the lib was already masked or disabled), but check to make sure that the service was stopped.

There are definitely some cases where you wind up with a stopped service that has not actually been masked, which is not great. I'll poke at systemd a bit and see if there's a more complete way of doing things ...

petevg: Cool! The underlying methods are idempotent. You can repeatedly mask or unmask, or enable or disable, something without systemctl throwing an error. I got rid of the tries in code.

jnsgruk (approved):
LGTM, thanks :)

cc @MiaAltieri: I think this gets you what you were looking for!

PR #39, opened 2022-02-10
remove empty string from args passed to _snap

### LINKED ISSUE #38: snap.ensure raises SnapError if channel is defined but confinement isn't
`snap.ensure(snap.SnapState.Latest, channel='edge')`
fails with:
```
charms.operator_libs_linux.v1.snap.SnapError: ('Could not %s snap [%s]: %s', ['snap', 'install', 'microsample', '', '--channel="edge"'], 'microsample', '')
```
on install hook.
note the empty string passed as optargs to _snap; that is the confinement.
Proposed solution: before _install calls _snap, we should filter out the '' from the args list.

Have a branch fixing this bug, will open a PR as soon as I have time.

PietroPasotti:
Fixes #38

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -271,16 +271,17 @@ class Snap(object):
           channel: the channel to install from
           cohort: optional, the key of a cohort that this snap belongs to
         """
-        confinement = "--classic" if self._confinement == "classic" else ""
-        channel = '--channel="{}"'.format(channel) if channel else ""
-        args = [confinement, channel]
+        cohort = cohort or self._cohort
 
-        if not cohort:
-            cohort = self._cohort
+        args = []
+        if self.confinement == "classic":
+            args.append("--classic")
+        if channel:
+            args.append('--channel="{}"'.format(channel))
         if cohort:
             args.append('--cohort="{}"'.format(cohort))
 
-        self._snap("install", filter(None, args))
+        self._snap("install", args)
 
     def _refresh(
         self,
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -81,7 +81,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 
 def _cache_init(func):
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -197,14 +197,10 @@ class Snap(object):
 
     def __eq__(self, other) -> bool:
         """Equality for comparison."""
-        return (
-            isinstance(other, self.__class__)
-            and (
-                self._name,
-                self._revision,
-            )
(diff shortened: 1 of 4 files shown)


PietroPasotti:
Not sure why I had to reformat two places I did not modify to satisfy the linter, did something change in the CI pipeline?

jnsgruk (approved):

pengale (changes requested):
IIRC, leaving in an empty string in the case of a missing argument is a pattern in the rest of this library.

I was a little worried about it, and wrote an integration test exercising snap.ensure, with empty string. See `test_snap_ensure` in `tests/integration/test_snap.py`.

While I don't think that it's necessarily a bad idea to remove the empty strings, just for the sake of being neat, it's strange that the integration test isn't failing, and I'm not sure that this PR addresses the bug. 

To feel confident that we're merging a bugfix here, I'd like to see a failing integration test that is fixed by these changes.

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -284,7 +280,7 @@ def _install(self, channel: Optional[str] = "", cohort: Optional[str] = "") -> N
         if cohort:
             args.append('--cohort="{}"'.format(cohort))
 
-        self._snap("install", args)
+        self._snap("install", filter(None, args))
```
rwcarlsen, line 283: I'd really rather we just built the args cleanly rather than be loose and the fix them immediately after.

rwcarlsen (changes requested):
As @pengale mentioned - a test would be nice to ensure we are fixing what we intend. Or at least look into that integration test that isn't failing and make sure nothing is wonky.  If this is definitely the fix we expect, then I support this change on the condition that we can be consistent with the other methods that call `_snap` - we should fix all cases of passing empty args.

lib/charms/operator_libs_linux/v1/snap.py
PietroPasotti: I'm with you there, was tempted by the low complexity of this solution; instead of burdening each 'should we add this arg' with conditional logic... 
are we sure that would read better?

PietroPasotti:
The issue is that the current integration tests attempt to `.ensure()` an already-installed snap, so my guess is the `snap install ...` command early-exits before it has time to parse all the subsequent arguments.
Adding a new test that `.ensure`s a not-already-installed snap, like `vlc`, reveals the error. Adding that to the integration tests.

image

pengale (approved):
+1. But see comments.

lib/charms/operator_libs_linux/v1/snap.py
rwcarlsen: `sed -i ""` is a thing.  So yes - technically possible for meaningful empty args to be used by commands in general.

rwcarlsen: One approach:

```python
args = []
if self._confinement == "classic":
    args.append('--classic')
if channel:
    args.append('--channel')
```

It's more lines, but dead simple.

rwcarlsen:
I've changed my mind merging this in without combing through all the snap commands/calls is fine.  But I would like to get away from using `filter` still.

jnsgruk:
One final change before merge: can you bump `LIBPATCH` then I'll merge and get it published. Thanks!

jnsgruk:
Merged and published to Charmhub

PR #41, opened 2022-03-03
Fix issues with apt methods (add_package/remove_package)

### LINKED ISSUE #40: `apt.add_package` ignores packages that are reported by `dpkg -l`, but not really installed.
> library: `charms.operator-libs-linux.v0.apt`
> version: current `main`, hash `c2e2f851daa62870239ae82faf40bfd0b7bee296`

A package that shows up in the output of `dpkg -l` is considered installed, even though it might not be. An example would be a situation when a package was removed, but not purged, and left config files behind:

```
root@juju-cf280c-xenial-0:~# dpkg -l ubuntu-advantage-tools
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name                              Version               Architecture          Description
+++-=================================-=====================-=====================-========================================================================
rc  ubuntu-advantage-tools            27.2.2~16.04.1        amd64                 management tools for Ubuntu Advantage
```

The code uses a regex to process this output, and it does extract a `package_status` [0] (which would be "rc" in this case), but it is not used. The library should check this field, note that the package is not really installed, and deal with the situation appropriately. Currently it is essentially a no-op as it considers the package already installed, even though it is not.

[0] https://github.com/canonical/operator-libs-linux/blob/main/lib/charms/operator_libs_linux/v0/apt.py#L413

jguedez:
* Use package status from dpkg to determine if a package is installed.
* Handle the case where no packages are removed.

Fixes #40.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -124,7 +124,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 6
+LIBPATCH = 7
 
 
 VALID_SOURCE_TYPES = ("deb", "deb-src")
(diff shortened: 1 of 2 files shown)


rbarry82 (approved):

pengale (approved):
This looks good to me. Thank you for the catch!

pengale:
Code merged, and updated library published!

PR #43, opened 2022-05-09
Updated "Raises" section in add_package() docstring to match method.

WRFitch:
This method actually raises TypeErrors and PackageErrors, depending on behaviour. This PR adds this information to the docstring.


jnsgruk:
Thanks @WRFitch!

PR #44, opened 2022-05-30
feat: add service management functionality to snap.py

marcoppenheimer:
## Motivation
- Add daemon service management functionality within `snapctl` alongside a `Snap` instance, keeping it contained with the popular `Snap` instance for charming

## Main Changes
- feat: add `start`, `stop`, `logs`, `restart` methods to `Snap` object (matching `snapctl`)

    - Methods can take a `list`,`str` or None services, evaluating to `snap start lxd.activate lxd.user-daemon`, `snap start lxd.user-daemon` or `snap start lxd`
    - These are called by `Snap._snap_daemons()` which (attempts) to ensure that a given snap instance can only launch services containing it's own `_name`
- feat: add `Snap.services`, `Snap.apps` properties to get metadata from the API for installed snaps

    - `Snap.apps` looks like `{"snap":"charmcraft","name":"charmcraft"},{"snap":"juju","name":"fetch-oci","daemon":"oneshot","daemon-scope":"system","enabled":true}`, which could be useful for a user to know what commands are facilitated by the installed snap
    - `Snap.services` looks like below. I felt it gave a useful format for accessing snap daemons by key, with the `snap` field removed due to it being redundant for a named `Snap` instance, and filling information missing from the API (i.e `active` would not exist in the API response if not active)
      ```
      {
        "fetch-oci": {
          "daemon": "oneshot",
          "daemon-scope": "system",
          "enabled": False,
          "active": False,
          "activators": [],
        },
      }
      ```

## Minor Changes
- style: set API request return type to `Any`, because JSON is JSON
- style: minor formatting changes
- fix: update some old `str.format` usages that I don't think worked before

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -62,15 +62,14 @@ import logging
 import os
 import socket
 import subprocess
-from subprocess import CompletedProcess
 import sys
 import urllib.error
 import urllib.parse
 import urllib.request
 from collections.abc import Mapping
 from enum import Enum
-from subprocess import CalledProcessError
-from typing import Dict, Iterable, List, Optional, Union
+from subprocess import CalledProcessError, CompletedProcess
+from typing import Any, Dict, Iterable, List, Optional, Union
 
 logger = logging.getLogger(__name__)
 
@@ -187,6 +186,7 @@ class Snap(object):
         channel: str,
         revision: str,
         confinement: str,
+        apps: Optional[List[Dict]] = [],
         cohort: Optional[str] = "",
     ) -> None:
         self._name = name
@@ -195,6 +195,7 @@ class Snap(object):
         self._revision = revision
         self._confinement = confinement
         self._cohort = cohort
+        self._apps = apps or []
 
     def __eq__(self, other) -> bool:
         """Equality for comparison."""
@@ -299,9 +300,7 @@ class Snap(object):
 
         self._snap_daemons("start", services, options)
 
-    def stop(
-        self, services: Optional[List[str]] = None, disable: Optional[bool] = False
-    ) -> None:
+    def stop(self, services: Optional[List[str]] = None, disable: Optional[bool] = False) -> None:
         """Stops a snap's services.
 
         Args:
@@ -434,7 +433,17 @@ class Snap(object):
                 # The snap is installed, but we are changing it (e.g., switching channels).
                 self._refresh(channel, cohort)
 
+        self._update_snap_apps()
         self._state = state
+
+    def _update_snap_apps(self) -> None:
+        """Updates a snap's apps after snap changes state."""
+        snap_client = SnapClient()
+        try:
+            self._apps = snap_client.get_installed_snap_apps(self._name)
+        except SnapAPIError:
+            logger.warning(f"Unable to retrieve snap apps for {self._name}")
+            self._apps = []
 
     @property
     def present(self) -> bool:
@@ -481,21 +490,27 @@ class Snap(object):
         return self._confinement
 
     @property
-    def services(self) -> dict:
-        """Returns (if any) the services of the snap."""
-        service_map = {}
-        try:
-            services = self._snap("services")
-            headers = [line.split() for line in services.splitlines()[:1]][0][1:]  # removing output headers 
-            services = [line.split() for line in services.splitlines()[1:]]
-            for item in services:
-                service_name = item.pop(0).split(".")[1]  # removing snap name from service name
-                item_map = {k: v for k, v in zip(headers, item)}
-                service_map[service_name] = item_map
-            return service_map
-        except SnapError as e:
-            logger.error(e)
-            return service_map
+    def apps(self) -> List:
+        """Returns (if any) the installed apps of the snap."""
+        self._update_snap_apps()
+        return self._apps
+
+    @property
+    def services(self) -> Dict:
+        """Returns (if any) the installed services of the snap."""
+        self._update_snap_apps()
+        services = {}
+        for app in self._apps:
+            if "daemon" in app:
+                services[app["name"]] = {
+                    "daemon": app["daemon"],
+                    "daemon-scope": app.get("daemon-scope", None),
+                    "enabled": app.get("enabled", False),
+                    "active": app.get("active", False),
+                    "activators": app.get("activators", []),
+                }
+
+        return services
 
 
 class _UnixSocketConnection(http.client.HTTPConnection):
@@ -576,7 +591,7 @@ class SnapClient:
         path: str,
         query: Dict = None,
         body: Dict = None,
-    ) -> Dict:
+    ) -> Any:
         """Make a JSON request to the Snapd server with the given HTTP method and path.
 
         If query dict is provided, it is encoded and appended as a query string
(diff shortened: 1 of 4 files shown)


lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -236,7 +238,28 @@ def _snap(self, command: str, optargs: Optional[Iterable[str]] = None) -> str:
         try:
             return subprocess.check_output(_cmd, universal_newlines=True)
         except CalledProcessError as e:
-            raise SnapError("Could not %s snap [%s]: %s", _cmd, self._name, e.output)
+            raise SnapError("Could not {} for snap [{}]: {}".format(_cmd, self._name, e.output))
+
+    def _snap_daemons(
```
PietroPasotti, line 243: I don't like the name. It reads like a getter. Add a verb for clarity: _snap_daemons_run?

Also why is it daemons plural? _snap_daemon_run?

Also why is there a daemon in there? is it interesting that it's a daemon? _snap_run?

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -264,6 +287,63 @@ def unset(self, key) -> str:
...
+
+        Args:
+            services (list): (optional) list of individual snap services to start (otherwise all)
+            enable (bool): (optional) flag to enable snap services on start. Default `false`
+        """
+        options = []
+        if enable:
+            options.append("--enable")
+
+        self._snap_daemons("start", services, options)
```
PietroPasotti, line 301: are '--enable' and '--disable' the only possible options one can pass to snap cmd? If not, why is it the only one we wrap? 
How does the next PR extend this pattern?

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -356,8 +436,18 @@ def ensure(
...
+        self._update_snap_apps()
         self._state = state
 
+    def _update_snap_apps(self) -> None:
+        """Updates a snap's apps after snap changes state."""
+        snap_client = SnapClient()
+        try:
+            self._apps = snap_client.get_installed_snap_apps(self._name)
+        except SnapAPIError:
+            logger.warning(f"Unable to retrieve snap apps for {self._name}")
```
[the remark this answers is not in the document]
PietroPasotti: I know, I forget it all the time as well.

tests/integration/test_snap.py
```diff
@@ -96,3 +98,51 @@ def test_snap_ensure():
...
+    assert kp.services["daemon"]["active"] is False
+    assert kp.services["daemon"]["enabled"] is False
+
+
+def test_snap_logs():
+    cache = snap.SnapCache()
+    kp = cache["kube-proxy"]
+    kp.ensure(snap.SnapState.Latest, classic=True, channel="latest/stable")
+
+    # Terrible means of populating logs
```
PietroPasotti, line 131: :D

PietroPasotti (changes requested):
Mostly nits, but a couple of points need a closer look. Looks great for the rest!

lib/charms/operator_libs_linux/v1/snap.py
marcoppenheimer: It was an attempt to leave a low footprint, as it's mostly the same command as `Snap._snap()` but for `daemons`.

I can rename the former to `Snap._snap_run()` and `Snap._snap_daemon_run()`, what do you think?

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -481,7 +594,7 @@ def _request(
         path: str,
         query: Dict = None,
         body: Dict = None,
-    ) -> Dict:
+    ) -> Any:
```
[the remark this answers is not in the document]
marcoppenheimer: Ha, I saw that while googling and thought it looked too hideous so went with `Any`, I'll update.

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -628,11 +746,12 @@ def _load_info(self, name) -> Snap:
...
-            SnapState.Available,
-            info["channel"],
-            info["revision"],
-            info["confinement"],
+            name=info["name"],
+            state=SnapState.Available,
+            channel=info["channel"],
+            revision=info["revision"],
+            confinement=info["confinement"],
+            apps=None,
```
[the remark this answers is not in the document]
marcoppenheimer: I remember why I did it now.

`Snap.__repr__` would print `apps`, so I wanted it populated at `__init__`, without having to make an API call for every Snap that gets initialised in `SnapCache._load_installed_snaps` and `SnapCache._load_info`.

Do you have an idea on how to avoid that?

lib/charms/operator_libs_linux/v1/snap.py
PietroPasotti: Mmm ok, yes. An idea to have less repetitions all over would be:
`cmd = ['cmd str', 'option str'] if option else ['cmd str']
`self._snap_daemons(cmd, services)`

instead of accepting `options=Iterable[str]` when in practice it's only ever going to be a single string (or None)

lib/charms/operator_libs_linux/v1/snap.py
PietroPasotti: Aha no then I'd say uniformity with existing code takes precedence, we have _snap, and this apparently is a daemon thing, so let's keep `_snap_daemon`. 

(I don't know much about snaps)

lib/charms/operator_libs_linux/v1/snap.py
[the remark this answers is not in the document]
marcoppenheimer: Excellent, I'll ignore `repr` and let it store after reference to `.apps`!

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -236,7 +272,30 @@ def _snap(self, command: str, optargs: Optional[Iterable[str]] = None) -> str:
...
+
+    def _snap_daemons(
+        self,
+        command: List[str],
+        services: Optional[List[str]] = None,
+    ) -> CompletedProcess:
+
+        if services:
+            # an attempt to keep the command constrained to the snap instance's services
+            services = [f"{self._name}.{service}" for service in services]
```
jnsgruk, line 289:
```suggestion
            services = ["{}.{}".format(self._name, service) for service in services]
```

jnsgruk (approved):
LGTM, modulo fixing the failing tests on Python 3.5+

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -93,6 +93,39 @@ def inner(*args, **kwargs):
...
+    """Data wrapper for snap services."""
+
+    def __init__(
+        self,
+        daemon: Optional[str] = None,
+        daemon_scope: Optional[str] = None,
+        enabled: bool = False,
+        active: bool = False,
+        activators: List[str] = [],
+        **kwargs,
```
jnsgruk, line 110:
```suggestion
        **kwargs
```

PietroPasotti (approved):

jnsgruk:
Merged and published the the store, if you run `charmcraft fetch-lib` you'll get the changes! :) Thanks!

PR #45, opened 2022-07-28
Adds ability to hold snap refreshes system wide

jnsgruk:
The feature to hold an individual snap is not yet in snapd, but this allows a charm author to hold snap refreshes system-wide


jnsgruk:
Reviewers note: the failing integration tests are unrelated to this change and appear to be related to the logs impl.

pengale (approved):
Can confirm that the test failure is unrelated (fwiw, all tests pass when run locally). 

This makes sense, and the interface is good. I am +1.

Issue #49, opened 2022-09-27
snap: `install_local()` fails for some kwarg combinations


mert-kirpici:
The module level function `install_local()` is failing to install the snap file when given `dangerous=True` only. This is due to an empty string being passed as an argument to the underlying `subprocess.check_output()` call, which the `snap` process does not really appreciate.

https://github.com/canonical/operator-libs-linux/blob/9d197d7306fcf28ce976e44bd01f4a9cf06c1985/lib/charms/operator_libs_linux/v1/snap.py#L900-L909

Here is a similar scenario:
```
miles@t440:/tmp$ python3
Python 3.8.10 (default, Jun 22 2022, 20:18:18) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from subprocess import check_output
>>> out = check_output(["sudo", "snap", "install", "/tmp/curl_1093.snap", "", "--dangerous"])
error: cannot install snap with empty name
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['sudo', 'snap', 'install', '/tmp/curl_1093.snap', '', '--dangerous']' returned non-zero exit status 1.
```

PR #50, opened 2022-09-27
snap: remove empty args from check_output

mert-kirpici:
There were two issues with the module level function install_local() that needed adressing.

- The unused function parameter: `self`
- Some combinations of the function parameters `classic` and `dangerous` ended up in passing empty arguments to the `snap` process when called with subprocess.check_output(), which it did not really like.

Fixes: #49
Signed-off-by: Mert Kırpıcı <mert.kirpici@canonical.com>


jnsgruk (approved):
Nice, thank you!

PR #53, opened 2022-10-10
passwd: create home dir for system users

mert-kirpici:
The `--system` argument needs also `--create-home` argument to be present for the creation of home directory for system users.

This change adds a new parameter to the `add_user` function signature, called `create_home` which defaults to True to provide backwards compatibility with the current api. After this change, home directory will be always created regardless of regular users or system users unless False is passed for `create_home`.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/passwd.py
+++ b/lib/charms/operator_libs_linux/v0/passwd.py
@@ -99,6 +99,7 @@ def add_user(
     secondary_groups: List[str] = None,
     uid: int = None,
     home_dir: str = None,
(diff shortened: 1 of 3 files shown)


PietroPasotti:
LGTM, thanks! 
After lint passes we can merge away.

mert-kirpici:
Hi @PietroPasotti thanks for the review. I changed the implementation. Decided a new function parameter would be more explicit. Updated the commit message and shamelessly increased the max-complexity for the linter... Could you review it again please? Sorry for the confusion

jnsgruk (approved):

Issue #54, opened 2022-11-18
snap charmlib's `install_local` tries to unpack the wrong line after installing package


NucciTheBoss:
I get this error when I use the `install_local` method in the snap charm library:

```python3
Traceback (most recent call last):
  File "/root/./install", line 126, in <module>
    holder._run()
  File "/root/./install", line 87, in _run
    self._handle_snap_install()
  File "/root/./install", line 118, in _handle_snap_install
    snap.install_local(
  File "/usr/local/lib/python3.10/dist-packages/cleantest/pkg/handler/snap_handler.py", line 798, in install_local
    snap_name, _ = result.split(" ", 1)
ValueError: not enough values to unpack (expected 2, got 1)
```

The the local snap file is successfully installed, but it fails to return the `Snap` wrapper class. Upon further inspection, line 910 seems to be the culprit:

https://github.com/canonical/operator-libs-linux/blob/69f7f59ed3707b01737e1fd0ef278193fefbea82/lib/charms/operator_libs_linux/v1/snap.py#L909-L917

Line 910 sets the value of `result` to an empty string, which causes the unpack to fail on line 911. This is because the first line captured by `check_output` is a blank line. I can tell that we are trying to retrieve the snap name from the captured output, so the index should be changed to `-1`. This will pull the last line that contains the snap name and version. It looks like `check_output` is throwing a wrench in the works because even though `snap install ...` is printing to the same line in stdout, `check_output` is registering each emission as a newline of captured output.

Issue #55, opened 2022-11-18
Snap API raises errors when searching the snap cache for locally installed snap package


NucciTheBoss:
During my quest to fix #54, I noticed that I would receive HTTP errors when contacting the `snapd.socket` after I installed the local snap package. This error prevented `install_local` from returning the `Snap` wrapper class.

Upon further investigation, I found the issue to be that `check_output` was also capturing ANSI escape sequences:

```python
>>> import subprocess
>>> res = subprocess.check_output(["snap", "install", "./local_amd64.snap", "--dangerous"], universal_newlines=True).splitlines()[-1]
>>> snap_name, _ = res.split(" ", 1)
>>> snap_name
'\x1b[0m\x1b[?25h\x1b[Khello-world-gtk'
```

These escape sequences need to be filtered out of the `snap_name` string for the snapd API call to function correctly.

PR #56, opened 2022-11-18
Fix snap charmlib's `install_local` method

NucciTheBoss:
Pull request contains fixes for the snap charm library's `install_local` method. Here is what I changed:

1. On `subprocess.check_output(...).splitlines()`, I changed the selected index from 0 to -1.
2. I added a regular expression that filters ANSI characters out from the captured `snap_name` so that the API request to the snapd socket does not fail.

Closes: #54, #55

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 1
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 
 # Regex to locate 7-bit C1 ANSI sequences
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -60,6 +60,7 @@ import http.client
 import json
 import logging
 import os
+import re
 import socket
 import subprocess
 import sys
@@ -907,8 +908,10 @@ def install_local(
     if dangerous:
         _cmd.append("--dangerous")
     try:
-        result = subprocess.check_output(_cmd, universal_newlines=True).splitlines()[0]
+        ansi_filter = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
+        result = subprocess.check_output(_cmd, universal_newlines=True).splitlines()[-1]
         snap_name, _ = result.split(" ", 1)
+        snap_name = ansi_filter.sub("", snap_name)
 
         c = SnapCache()
 
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -86,6 +86,10 @@ LIBAPI = 1
 LIBPATCH = 5
 
 
+# Regex to locate 7-bit C1 ANSI sequences
+ansi_filter = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
+
+
 def _cache_init(func):
     def inner(*args, **kwargs):
         if _Cache.cache is None:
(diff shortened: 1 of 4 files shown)


NucciTheBoss:
I did not make any changes to the `logs` functionality of the charm library.

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -907,8 +908,10 @@ def install_local(
     if dangerous:
         _cmd.append("--dangerous")
     try:
-        result = subprocess.check_output(_cmd, universal_newlines=True).splitlines()[0]
+        ansi_filter = re.compile(r"\x1B(?:[@-Z\\-_]|\[[0-?]*[ -/]*[@-~])")
```
PietroPasotti, line 911: I'd make this a global (or a class attr) so as to only compile it once per run/import

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -907,8 +908,10 @@ def install_local(
...
+        result = subprocess.check_output(_cmd, universal_newlines=True).splitlines()[-1]
```
PietroPasotti, line 912: didn't know that. Cool.

lib/charms/operator_libs_linux/v1/snap.py
PietroPasotti, line 911: How did you pick that regex? If I scan through https://stackoverflow.com/questions/14693701/how-can-i-remove-the-ansi-escape-sequences-from-a-string-in-python I find tons of answers including 

```
ansi_regex = r'\x1b(' \
             r'(\[\??\d+[hl])|' \
             r'([=<>a-kzNM78])|' \
             r'([\(\)][a-b0-2])|' \
             r'(\[\d{0,2}[ma-dgkjqi])|' \
             r'(\[\d+;\d+[hfy]?)|' \
             r'(\[;?[hf])|' \
             r'(#[3-68])|' \
             r'([01356]n)|' \
             r'(O[mlnp-z]?)|' \
             r'(/Z)|' \
             r'(\d+)|' \
             r'(\[\?\d;\d0c)|' \
             r'(\d;\dR))'
```

which looks very different.
Maybe add a comment explaining your choice?
And perhaps a couple of tests :)

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -907,8 +908,10 @@ def install_local(
...
         snap_name, _ = result.split(" ", 1)
+        snap_name = ansi_filter.sub("", snap_name)
```
PietroPasotti, line 914: A more general comment: it took you some debugging to pinpoint the failing line and the assumptions that didn't hold. How about wrapping some of the assumptions you're making here (the 'getitems', etc...) in try/except blocks, to raise some helpful exceptions and log some variables?

PietroPasotti (changes requested):
Thanks for taking the time to pinpoint this issue, I think we need a bit more testing and a couple of comments before we can merge, but it's definitely going in the right direction :)

lib/charms/operator_libs_linux/v1/snap.py
NucciTheBoss: There are quite a few possible expressions for filtering out ANSI escape sequences. The expression that I am using only filters out 7-bit C1 ANSI sequences. If I found the answer that contains the regular expression copied above, the expression does not actually work. It incompletely filters out the escape sequences; it does not remove certain parts of the ANSI escape sequences from the string.

I'll add a comment that says the regular expression removes 7-bit C1 ANSI escape sequences, and I'll see about adding some tests that filter the ANSI sequences out of captured output from `snap install ...`.

lib/charms/operator_libs_linux/v1/snap.py
NucciTheBoss: Sounds like a good idea. I could probably add a top-level try-expect block that captures Snap API errors. We won't need to worry about the unpacking since the captured output piped to a list is being sliced correctly now, but we should make sure that the name is formatted correctly (i.e. no ANSI escape sequences).

NucciTheBoss: Added try-except block in https://github.com/canonical/operator-libs-linux/pull/56/commits/8b13e63cdd98108dfd70e978ad4db8a225985b94.

lib/charms/operator_libs_linux/v1/snap.py
NucciTheBoss: Made global in https://github.com/canonical/operator-libs-linux/pull/56/commits/8b13e63cdd98108dfd70e978ad4db8a225985b94.

NucciTheBoss: Added tests in https://github.com/canonical/operator-libs-linux/pull/56/commits/39ab84cc5baf45aef7a799849cdefe61ef181b31 and I added a comment explaining what the regular expression does in https://github.com/canonical/operator-libs-linux/pull/56/commits/8b13e63cdd98108dfd70e978ad4db8a225985b94. I needed to push a extra commit however because the linter was not happy with the formatting of the test I added.

jnsgruk (approved):

jnsgruk:
You just need to bump LIBPATCH, then this will be good to merge/publish.

NucciTheBoss:
Bumped the LIBPATCH version. Seems like the CI jobs are having issues. For example, the Python 3.5 job could not pull some of its dependencies. Everything should be good to go!

jnsgruk:
Merged and published. Thanks @NucciTheBoss!

Issue #62, opened 2023-01-12
snap ensure hangs


MiaAltieri:
## Problem
On occasion running `snap_package.ensure(snap.SnapState.Latest, channel=snap_channel)`, this line of code will hang for anywhere between several minutes to several hours before erroring out. 

## Context
This behaviour has not been experienced on the commandline only with this library by both @MiaAltieri and @taurus-forever 

After looking into the code it appears to be an issue with [check_output](https://github.com/canonical/operator-libs-linux/blob/46d687651373fe0b57ef2947f4f162c0e637c865/lib/charms/operator_libs_linux/v1/snap.py#L279) failing but no useful error message is given:
```
unit-mongodb-2: 16:12:54 ERROR unit.mongodb/2.juju-log An exception occurred when installing percona-backup-mongodb. Reason: Snap: 'percona-backup-mongodb'; command ['snap', 'install', 'percona-backup-mongodb', '--channel="edge"'] failed with output = ''
```

## Proposed Solution 
Add some sort of timeout+retry for ensure that the snap can be installed in a timely manner.

MiaAltieri:
I will work on a solution for this using tenacity and the `timeout` parameter

PR #63, opened 2023-01-12
add timeout to snap operations

MiaAltieri:
solves issue #62

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -72,13 +72,8 @@ from datetime import datetime, timedelta, timezone
 from enum import Enum
 from subprocess import CalledProcessError, CompletedProcess, TimeoutExpired
 from typing import Any, Dict, Iterable, List, Optional, Union
-from tenacity import (
-    before_log,
-    retry,
-    stop_after_attempt,
-    wait_fixed,
-)
 
+from tenacity import before_log, retry, stop_after_attempt, wait_fixed
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -70,8 +70,15 @@ import urllib.request
 from collections.abc import Mapping
 from datetime import datetime, timedelta, timezone
 from enum import Enum
-from subprocess import CalledProcessError, CompletedProcess
+from subprocess import CalledProcessError, CompletedProcess, TimeoutExpired
 from typing import Any, Dict, Iterable, List, Optional, Union
+from tenacity import (
+    before_log,
+    retry,
+    stop_after_attempt,
+    wait_fixed,
+)
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -70,9 +70,11 @@
 from collections.abc import Mapping
 from datetime import datetime, timedelta, timezone
 from enum import Enum
-from subprocess import CalledProcessError, CompletedProcess
+from subprocess import CalledProcessError, CompletedProcess, TimeoutExpired
 from typing import Any, Dict, Iterable, List, Optional, Union
 
+from tenacity import before_log, retry, stop_after_attempt, wait_fixed
```
benhoyt, line 76: I don't think we (normally) want to introduce a dependency on a 3rd party library inside a charm lib, right? If we do, in addition to adding the charm lib, you'd have to manually add the dependency to the charm's requirements.txt to satisfy it. Doesn't seem worth it for something as straight-forward as a retry.

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -262,21 +264,35 @@ def __str__(self):
             str(self._state),
         )
 
-    def _snap(self, command: str, optargs: Optional[Iterable[str]] = None) -> str:
+    @retry(
+        stop=stop_after_attempt(5),
+        wait=wait_fixed(5),
+        reraise=True,
+        before=before_log(logger, logging.DEBUG),
+    )
```
benhoyt, line 272: My concern with adding retries is that we may be somewhat blindly patching over an issue that we don't know the cause of.

I think part of the issue is we're not showing the stderr from the `subprocess.check_output()` call, only the stderr, which isn't very useful. The error information from snap will be in stderr, and stdout will be empty. So I think we should change the `check_output()` block below to the following:

```python
try:
    return subprocess.check_output(_cmd, universal_newlines=True, stderr=subprocess.PIPE)
    raise SnapError(
        "Snap: {!r}; command {!r} failed with stderr = {!r}".format(
            self._name, _cmd, e.stderr
        )
    )
```

The above change should also be done in the `check_output()` call in `install_local` and its exception handling, and in `_system_set` below that:

```python
    try:
        subprocess.check_call(_cmd, universal_newlines=True, stderr=subprocess.PIPE)
        raise SnapError("Failed setting system config '{}' to '{}': {}".format(config_item, value, e.stderr))
```

I don't think we should add the timeout or the retries at this point, until we know more.

benhoyt: Would be good to add unit tests that check the error case and assert that the stderr is in the exception message. Maybe by patching `subprocess.check_output` with a version that instead calls `check_output(['sleep', 'x'])` or something, so it's slightly realistic?

benhoyt (changes requested):
Hi @MiaAltieri - thanks for the PR! I don't think we should merge this with the added dependency and the retries as is -- I've added comments with a few things to consider here.

lib/charms/operator_libs_linux/v1/snap.py
MiaAltieri: That makes sense @benhoyt  :)

lib/charms/operator_libs_linux/v1/snap.py
MiaAltieri: We could remove `timeout` for the time being until we figure out what the issue is here. (But, debugging this will take significant time, especially considering that without the `timeout` it can take up to several hours for `subprocess.check_call` to finish.)

 I have limited time and this issue is a blocker for my team. I will see if I have time to debug this issue.

PR #69, opened 2023-01-23
fix: make placeholder charm a machine charm

### LINKED ISSUE #47: Placeholder charm should be a machine charm
Currently, the placeholder charm that hosts these libraries has a default metadata.yaml, which references a placeholder container. This leads to the store to believe that the charm is a k8s charm.

And that is a little confusing :-)

This issue can be closed once the placeholder resource is removed from the metadata.

benhoyt:
These are machine-charm utilities, so make the placeholder charm a machine charm so it shows up correctly in Charmhub. Do this by removing containers / resources from metadata.yaml.

Also:

- Update Ubuntu version to 22.04 (not that it matters for charm libs)
- Don't need to be on the latest master "ops", just use latest release

Fixes #47

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -5,11 +5,7 @@ type: charm
 bases:
   - build-on:
       - name: "ubuntu"
-        channel: "20.04"
+        channel: "22.04"
     run-on:
       - name: "ubuntu"
-        channel: "20.04"
-parts:
-  charm:
-    build-packages:
-      - git
+        channel: "22.04"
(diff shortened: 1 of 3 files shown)


jnsgruk:
Ah nice, this is a good idea.

A thought I had for the future: it would be nice if deploying the charm did *something useful*. One idea that I had was to deploy a very simple web server that could render the docs for each lib. Not exactly high priority, but could be interesting - I imagine we could re-use the code from charmhub.io that renders the library docs pages.

jnsgruk (approved):

Issue #72, opened 2023-02-21
DebianRepository.import_key() should be a staticmethod so it can be called without a DebianRepository instance


toabctl:
My usecase is, that I want to import a GPG key for apt (eg. https://repos.influxdata.com/influxdata-archive_compat.key). For that, I only need that file and a filename how this key should be named. But that is currently not possible because I first need to create a `DebianRepository` instance which requires a couple of arguments which are not needed at all for my usecase.
So what I would like to do is:
```
key = """
-----BEGIN PGP PUBLIC KEY BLOCK-----
mQINBGPIEycBEACpG4qSjhxA6fh4QJVJxFVBvCFt9tVx/hDbKH0Ryy9iilyMeReC
...
-----END PGP PUBLIC KEY BLOCK-----
"""
DebianRepository.import_key(key)
```

benhoyt:
Hi @toabctl -- thanks for this. I've looked at your PR and I don't think we should change that to a static method, because as you've seen, right now it sets `self._gpg_key_filename` so this change wouldn't be backwards-compatible. I suggest one of two approaches here:

1) Either just bite the bullet and instantiate a `DebianRepository` with the 5 non-optional parameters, or create a dummy one using `DebianRepository.from_repo_line`. This would be the simplest way forward. Or:
2) Create a PR to refactor `DebianRepository.import_key` to a top-level `import_key` function that returns the file name it has written out, and then update `DebianRepository.import_key` to be defined as `self._gpg_key_filename = import_key(key)`.

Or we could factor out it out to a `read_key(key)` "side effect free" top-level function which returns a `(key_name, key_material)` tuple and then `DebianRepository.import_key` calls that and does the write. That might be a bit clumsier though as you'd still have to write it out.

**On a related note:** is there an existing bug in the `else` branch of this function? I see there it calls `self._write_apt_gpg_keyfile(key_name=key, key_material=key_gpg)` with `key_name=key` instead of `key_name=self._gpg_key_filename`, which doesn't seem right. Agreed? If so, you'd be welcome to submit a PR to fix this (ideally with tests of both branches :-).

toabctl:
> Hi @toabctl -- thanks for this. I've looked at your PR and I don't think we should change that to a static method, because as you've seen, right now it sets `self._gpg_key_filename` so this change wouldn't be backwards-compatible. I suggest one of two approaches here:
> 
[...]

PR #73, opened 2023-02-21
Make import_key a stand-alone function for use with a DebianRepository

toabctl:
This PR allows users to import a key without a dummy `DebianRepository` instance, by adding a top-level `import_key` function that does the work, and then refactoring `DebianRepository.import_key` to use this. Users that want to import a key but don't have a `DebianRepository` can just call the new top-level function.


toabctl:
I know this breaks API. I just created it to provide a POC for #72

benhoyt (approved):
Thanks for that -- looks good. I also bumped up LIBPATCH and made the key_name fix.

PR #74, opened 2023-02-23
Bump LIBPATCH for two libraries that had been modified but not bumped

benhoyt:
These had been modified (systemd only with comment changes, but still) in previous PRs but LIBPATCH hadn't been bumped, so `charmhub publish-lib` was complaining that the content was different.


benhoyt:
Merging without further review as these LIBPATCH versions seem clearly out of date.

PR #78, opened 2023-03-09
build: update to use ruff/codespell

### LINKED ISSUE #77: Upgrade lint tooling in line with styleguide
Other charms and repos have moved away from `flake8` / `isort` to `ruff` - we should follow suit.

jnsgruk:
There looks like there are lots of changes here, but the key is:

- Update from black/isort -> black/ruff
- Introduce codespell

Then there are a bunch of changes to docstrings and spellings to conform to ruff's rules. Main change is making first line of docstrings use the imperative mood.

Fixes #77


jnsgruk:
Note that I've not bumped the LIBPATCH for the libraries, but when #79 gets merged, the versions are bumped and the libs should release automatically.

benhoyt (approved):
This looks good to me, thank you!

Issue #80, opened 2023-03-15
Should `snap.hold_refresh` change it's functionality to allow for infinite holds?


zmraul:
Currently setting `snap.hold_refresh(days=0)` will remove the hold on the snaps. What value would be reasonable to allow for indefinite holds?

benhoyt:
I don't think there is a way currently, as the `hold_refresh` function checks that days is less than or equal to 90 and raises a `ValueError` if not. However, as of Aug 2022 ([this snapd commit](https://github.com/snapcore/snapd/commit/e226d30a3c85ae90f1fe43f8d81ceb00f0dc7f4b)), snap allows the `refresh.hold` config option to be set to "forever" (a user can do this with `snap refresh --hold` or `sudo snap set system refresh.hold=forever`) to allow indefinite holds.

I'd be in favour of a PR that allows `hold_refresh(forever=True)` option that enabled this. @zmraul Would you be willing to put up a PR with tests?

I though about `hold_refresh(days=None)` but I think it's a bit confusing that it would allow `days=0` to mean no hold and `days=None` to mean indefinite (especially when both 0 and None are falsey in Python).

jnsgruk:
Agree with the proposal for `hold_refresh(forever=True)` - there are some other bits that landed recently like snap install by revision too

jnsgruk:
(I should also mention that a really important feature is the **per snap** hold you can now do - we should think about implementing that)

PR #81, opened 2023-03-20
feat: add forever hold to snap refresh.hold

zmraul:
Based on the feedback on #80, adds an option to hold the refresh of snaps forever.

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -976,15 +976,21 @@ def _system_set(config_item: str, value: str) -> None:
         raise SnapError("Failed setting system config '{}' to '{}'".format(config_item, value))
 
 
-def hold_refresh(days: int = 90) -> bool:
+def hold_refresh(days: int = 90, forever: bool = False) -> bool:
     """Set the system-wide snap refresh hold.
 
     Args:
         days: number of days to hold system refreshes for. Maximum 90. Set to zero to remove hold.
+        forever: if set  to True, will set a hold forever.
     """
+    if not isinstance(forever, bool):
+        raise ValueError("forever must be a boolean")
     # Currently the snap daemon can only hold for a maximum of 90 days
-    if not isinstance(days, int) or days > 90:
+    elif not isinstance(days, int) or days > 90:
         raise ValueError("days must be an int between 1 and 90")
+    elif forever:
+        _system_set("refresh.hold", "forever")
+        logger.info("Set system-wide snap refresh hold to: forever")
     elif days == 0:
         _system_set("refresh.hold", "")
         logger.info("Removed system-wide snap refresh hold")
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -981,20 +981,22 @@ def hold_refresh(days: int = 90, forever: bool = False) -> bool:
 
     Args:
         days: number of days to hold system refreshes for. Maximum 90. Set to zero to remove hold.
-        forever: if set  to True, will set a hold forever.
+        forever: if True, will set a hold forever.
     """
     if not isinstance(forever, bool):
-        raise ValueError("forever must be a boolean")
-    # Currently the snap daemon can only hold for a maximum of 90 days
-    elif not isinstance(days, int) or days > 90:
-        raise ValueError("days must be an int between 1 and 90")
-    elif forever:
+        raise TypeError("forever must be a bool")
+    if not isinstance(days, int):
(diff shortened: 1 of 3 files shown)


jnsgruk (approved):
Looks tidy, thanks!

jnsgruk:
Note that I think in general we should opt for a per-snap hold, especially when "forever holding" - so it's likely that we'll want to land that change before this is used too widely.

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -976,15 +976,21 @@ def _system_set(config_item: str, value: str) -> None:
         raise SnapError("Failed setting system config '{}' to '{}'".format(config_item, value))
 
 
-def hold_refresh(days: int = 90) -> bool:
+def hold_refresh(days: int = 90, forever: bool = False) -> bool:
     """Set the system-wide snap refresh hold.
 
     Args:
         days: number of days to hold system refreshes for. Maximum 90. Set to zero to remove hold.
+        forever: if set  to True, will set a hold forever.
```
benhoyt, line 984:
```suggestion
        forever: if True, will set a hold forever.
```

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -976,15 +976,21 @@ def _system_set(config_item: str, value: str) -> None:
...
     """
+    if not isinstance(forever, bool):
+        raise ValueError("forever must be a boolean")
```
benhoyt, line 987: Personally I'd don't think the runtime type check here will add anything. But if you want to keep it, this should be a TypeError, not a ValueError. And we should probably say "bool" rather than "boolean" to match the Python type name.

```suggestion
        raise TypeError("forever must be a bool")
```

lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -976,15 +976,21 @@ def _system_set(config_item: str, value: str) -> None:
...
     # Currently the snap daemon can only hold for a maximum of 90 days
-    if not isinstance(days, int) or days > 90:
+    elif not isinstance(days, int) or days > 90:
```
benhoyt, line 989: No need for `elif` as the above will have raised (I find it more confusing to keep track of when there's a run of "else"s.) Also, I think we should adjust the type check to raise `TypeError`, and the range check to also check for <0. My suggestion is that we simplify by doing the type checking up-front, and then the range checking where it's relevant:

```python
    if not isinstance(days, int):
        raise TypeError("days must be an int")
    if forever:
        ...
    elif days == 0:
        ...
    else:
        if not 1 <= days <= 90:
            raise ValueError("days must be between 1 and 90")
        ...
```

tests/integration/test_snap.py
```diff
@@ -167,6 +167,12 @@ def test_hold_refresh():
     assert f"hold: {hold_date}" in result.decode()
 
 
+def test_forever_hold_refresh():
+    snap.hold_refresh(forever=True)
+    result = check_output(["snap", "refresh", "--time"])
+    assert "hold: 2315-06-30" in result.decode()
```
benhoyt, line 173: What's this magical date? Will it change as time progresses, or is it a fixed date? On my system it's "2315-07-01", maybe due to me being on NZDT, or because it's already tomorrow for me? :-) If it is a fixed date, I recommend we do `assert "hold: 2315" in result.decode()`, if not, perhaps we just check with `snap get system refresh.hold`?

tests/unit/test_snap.py
```diff
@@ -534,6 +534,12 @@ def test_hold_refresh_invalid_non_int(self):
         except ValueError as e:
             self.assertEqual(str(e), "days must be an int between 1 and 90")
 
+    def test_hold_refresh_invalid_non_bool(self):
+        try:
+            snap.hold_refresh(forever="foobar")
+        except ValueError as e:
+            self.assertEqual(str(e), "forever must be a boolean")
```
benhoyt, line 541: When writing tests like this you have to assert that it did raise an error, rather than just "if it raises an error, check something". You do that like so (also note the `TypeError`):

```python
    snap.hold_refresh(forever="foobar")
```

I also suggest we *don't* check the string message, as that may change and isn't part of the contract we should be testing.

benhoyt (changes requested):
Looks good. I few minor suggestions.

And I agree with Jon -- would you be willing to follow up with a PR to add per-snap holds?

zmraul:
> And I agree with Jon -- would you be willing to follow up with a PR to add per-snap holds?

I can add an issue to solve that and take a look in the future. But I think it's outside of the scope of this PR and the needs we had for doing this.

PR #84, opened 2023-03-23
is-failed function for systemd services

Mehdi-Bendriss:
### Context:
When dealing with services, it is sometimes not enough to only know if a service is running or not, it is useful to know if the said service is in a failed state. 
This PR adds the `is-failed` status check.


jnsgruk (approved):
Looks good to me, thanks @Mehdi-Bendriss

PR #88, opened 2023-03-31
feat: install snap by revision

shayancanonical:
## Issue
We are now able to install snaps by revisions, but the snap charm library does not yet support this flag.

We would like to add support for the `--revision` flag so that we can pin our snap versions in charms.

## Solution
Add optional kwarg `revision` to the `ensure`, `_install` and `_refresh` methods of Snap in the snap charm lib

### DIFF
--- a/lib/charms/operator_libs_linux/v1/snap.py
+++ b/lib/charms/operator_libs_linux/v1/snap.py
@@ -222,7 +222,7 @@ class Snap(object):
         name,
         state: SnapState,
         channel: str,
-        revision: str,
+        revision: int,
         confinement: str,
         apps: Optional[List[Dict[str, str]]] = None,
         cohort: Optional[str] = "",
@@ -424,7 +424,10 @@ class Snap(object):
         self._snap_daemons(args, services)
 
     def _install(
-        self, channel: Optional[str] = "", cohort: Optional[str] = "", revision: Optional[str] = ""
+        self,
+        channel: Optional[str] = "",
+        cohort: Optional[str] = "",
+        revision: Optional[int] = None,
     ) -> None:
         """Add a snap to the system.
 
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v1/snap.py
```diff
@@ -423,12 +423,15 @@ def restart(
         args = ["restart", "--reload"] if reload else ["restart"]
         self._snap_daemons(args, services)
 
-    def _install(self, channel: Optional[str] = "", cohort: Optional[str] = "") -> None:
+    def _install(
+        self, channel: Optional[str] = "", cohort: Optional[str] = "", revision: Optional[str] = ""
```
jnsgruk, line 427: Strikes me that revision should be `int` throughout?

jnsgruk (changes requested):
A minor data type nit, and two things to consider:

- How does revision interact with channel when both are specified? 
- This needs both unit and integration tests before merging

Otherwise thanks for the contribution!

jnsgruk:
Gonna take a look at this later, @PietroPasotti and @marcoppenheimer could you also cast an eye over this?

shayancanonical:
Re specifying both `revision` and `channel`: I tested locally, and discovered that `revision` always takes precedence when both are specified. I was leaning towards keeping the lib lean and allowing users to specify both (and let the underlying `snap install` command determine how to resolve a mismatch). However, I am also open to adding validation to this lib to ensure both are not specified simultaneously

jnsgruk:
In that case I'd leave the lib clean, but update the doc strings to make the behaviour documented at least

marcoppenheimer (approved):
LGTM, no issues that I can see! Solid implementation 👍🏾

jnsgruk (approved):
LGTM, thanks.

Looks like the docstrings need some work for consistency in this lib (around wording/convention for optionality etc. Will open a separate issue)

PR #92, opened 2023-05-10
fix(snap): handle revision as str

### LINKED ISSUE #91: SnapCache does not handle dev installed snaps
I do not know if it's by design but `SnapCache` does not handle snaps when they are installed from a `.snap` file.

```
unit-openstack-hypervisor-0: 10:50:21 ERROR unit.openstack-hypervisor/0.juju-log certificates:4: Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-openstack-hypervisor-0/charm/venv/ops_sunbeam/guard.py", line 91, in guard
    yield
  File "/var/lib/juju/agents/unit-openstack-hypervisor-0/charm/venv/ops_sunbeam/charm.py", line 267, in configure_charm
    self.configure_unit(event)
  File "/var/lib/juju/agents/unit-openstack-hypervisor-0/charm/./src/charm.py", line 181, in configure_unit
    self.ensure_snap_present()
  File "/var/lib/juju/agents/unit-openstack-hypervisor-0/charm/./src/charm.py", line 168, in ensure_snap_present
    cache = snap.SnapCache()
  File "/var/lib/juju/agents/unit-openstack-hypervisor-0/charm/lib/charms/operator_libs_linux/v1/snap.py", line 774, in __init__
    self._load_installed_snaps()
  File "/var/lib/juju/agents/unit-openstack-hypervisor-0/charm/lib/charms/operator_libs_linux/v1/snap.py", line 831, in _load_installed_snaps
    revision=int(i["revision"]),
ValueError: invalid literal for int() with base 10: 'x1'

```

https://github.com/canonical/operator-libs-linux/blob/8f1b16899d12acec9d8cc876a1eca7731e802f17/lib/charms/operator_libs_linux/v1/snap.py#LL831C21-L831C21

gboutry:
Fix #91

Snap revisions can be strings and should be handled as strings

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -79,11 +79,11 @@ logger = logging.getLogger(__name__)
 LIBID = "05394e5893f94f2d90feb7cbe6b633cd"
 
 # Increment this major API version when introducing breaking changes
-LIBAPI = 1
+LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 0
 
 
 # Regex to locate 7-bit C1 ANSI sequences
@@ -222,7 +222,7 @@ class Snap(object):
         name,
         state: SnapState,
         channel: str,
-        revision: int,
+        revision: str,
         confinement: str,
         apps: Optional[List[Dict[str, str]]] = None,
         cohort: Optional[str] = "",
@@ -427,7 +427,7 @@ class Snap(object):
         self,
         channel: Optional[str] = "",
         cohort: Optional[str] = "",
-        revision: Optional[int] = None,
+        revision: Optional[str] = None,
     ) -> None:
         """Add a snap to the system.
 
@@ -454,7 +454,7 @@ class Snap(object):
         self,
         channel: Optional[str] = "",
         cohort: Optional[str] = "",
-        revision: Optional[int] = None,
+        revision: Optional[str] = None,
         leave_cohort: Optional[bool] = False,
     ) -> None:
         """Refresh a snap.
@@ -498,7 +498,7 @@ class Snap(object):
         classic: Optional[bool] = False,
         channel: Optional[str] = "",
         cohort: Optional[str] = "",
-        revision: Optional[int] = None,
+        revision: Optional[str] = None,
     ):
         """Ensure that a snap is in a given state.
 
@@ -575,7 +575,7 @@ class Snap(object):
         self._state = state
 
     @property
-    def revision(self) -> int:
+    def revision(self) -> str:
         """Returns the revision for a snap."""
         return self._revision
 
@@ -828,7 +828,7 @@ class SnapCache(Mapping):
                 name=i["name"],
                 state=SnapState.Latest,
                 channel=i["channel"],
-                revision=int(i["revision"]),
+                revision=i["revision"],
                 confinement=i["confinement"],
                 apps=i.get("apps", None),
             )
@@ -846,7 +846,7 @@ class SnapCache(Mapping):
             name=info["name"],
             state=SnapState.Available,
             channel=info["channel"],
-            revision=int(info["revision"]),
+            revision=info["revision"],
             confinement=info["confinement"],
             apps=None,
         )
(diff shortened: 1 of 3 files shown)


gboutry:
The `ensure` method could be modified to only accepted `int` as revisions to enforce in the public api the installation of valid revisions.

benhoyt:
I don't think this can go in as is, because it's a breaking change to change this field from an `int` to a `str`. So it would have to be a v2 of the charm lib if anything. But I'm not sure it's worth a v2 just for that change.

What's an "x" version? Does the "x" prefix mean local snap? Is the number after the "x" always 1, or does it increment? One possibility, especially if it's always "x1", would be to make "x1" map to version integer `-1`. I'm not sure what's best here -- I'd be interested to learn more about what the "x" means and when it happens.

gboutry:
That's what I thought at first, but it was a str before, and the library version went from 11 to 12 with this change (meaning that at version 11 there was no issue with dev versions like x1)

gboutry:
I'm not against keeping the external facing API's as int, but currently this library is making a charm fail because another snap installed on the host has `x1` as version. (It's not the snap that the charm is managing)

benhoyt:
Is a "dev version" a local `.snap` file? Are they always labelled "x1", or are there "x2" and "x3" and so on as well? What do you think of the idea to number dev versions with revision -1?

gboutry:
Yeah by what I meant with "dev version" is local revision. That would solve the problem, assuming we don't want to be able to address local revisions through the library

benhoyt:
Hmm, I'm not sure what's best to do here. Do we need to be able to distinguish different local revisions such as x1 and x2? If not, I lean towards parsing revision "xN" to -1. Alternatively, for backwards compatibility, we could make the revision either an int or a str: it would be an int normally, but for xN revisions it would be left as a str. Or we could bump the lib to v2 and change it to str (but do revisions need to be compared numerically?).

It was in https://github.com/canonical/operator-libs-linux/pull/88 where we changed from str to int previously (in a minor version, which doesn't seem great to me, as ints and strs behave very differently -- though clearly it hasn't changed caused major problems).

So, summarizing the alternatives:

1) Parse local revisions with the "x" prefix as the integer -1.
2) Make revision an int unless it's an "xN" revision, in which case make it a str.
3) Make the revision always a str again, and bump the lib to v2.

I lean to number 1 to keep things simple. Thoughts, @jnsgruk?

gboutry:
I don't believe we would want to manage local revision through a charm (there might be use case I don't see), so 1 is ok with me

wolsen:
Currently, if a snap is installed locally (e.g. a dev snap) it will definitely have the xN number per the [snap revision](https://github.com/snapcore/snapd/blob/5aeeeae239385693c7a9e5524f7ef4e23595e0f6/snap/revision.go#L33). It's clearly a string that is accepted, though it can be expected to be an integer for any snaps installed from the store.

It follows the specific path that @benhoyt suggested, if less than 0 then return x%d. Ben's suggestion is almost literally do the opposite: if start with x, then return -1. If we go that way, we could simply reverse the logic and do the opposite. However, I think in reality, the correct data type is a string as that is what the snap API is returning (and clearly it returns non-integers). While it is an integer in most cases (today), there's potential for the snap api/store api to change what the revision actually is (i.e. change from an int to a hash, etc). 

As such, I think we should just bump the api version and change the data type back to the string type. Arguably, since the major version had it as a string originally and it was changed in a minor - it could be just switched back at a minor level as well. I don't think that's good practice though as it *may* be a breaking API change for consumers.

*(as an aside, I did find it amusing that the integer is quoted like a string for the snap command line that ultimately gets executed)*

benhoyt:
Thanks @wolsen. Yeah, I think you're instinct is probably right. It is a string in the snapd API, so we should stick with that. 

@gboutry Can you please update this PR to basically be what you have, but do so in a v2 version of the lib? I don't think this kind of change should have been done in a minor bump before, and two wrongs don't make a right. :-)

gboutry:
Thanks @wolsen, @benhoyt

I updated to v2

benhoyt (approved):
This looks good to me - thanks!

gboutry:
Hi @benhoyt, do you have an idea when we can get this published to charmhub ? thank you for you response

jnsgruk:
@gboutry I just published this manually - it looks like our CI couldn't handle changing the major version.

@PietroPasotti I think you wrote the original lib publishing workflow, it seems [here](https://github.com/canonical/operator-libs-linux/actions/runs/4999440064/jobs/8955830350#step:3:72) it didn't pick up the major version bump.

PR #94, opened 2023-06-22
feat: create alias for a snap

dragomirp:
## Issue
We would like to be able to create snap aliases using the charm library

## Solution
Add `alias` method to create aliases.

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -273,10 +273,7 @@ class Snap(object):
           SnapError if there is a problem encountered
         """
         optargs = optargs or []
-        if command == "alias":
-            _cmd = ["snap", command, *optargs]
-        else:
-            _cmd = ["snap", command, self._name, *optargs]
+        _cmd = ["snap", command, self._name, *optargs]
         try:
             return subprocess.check_output(_cmd, universal_newlines=True)
         except CalledProcessError as e:
@@ -412,16 +409,24 @@ class Snap(object):
         """Remove the refresh hold of a snap."""
         self._snap("refresh", ["--unhold"])
 
-    def alias(self, application: str, alias: str = None) -> None:
+    def alias(self, application: str, alias: Optional[str] = None) -> None:
         """Create an alias for a given application.
 
         Args:
             application: application to get an alias.
             alias: (optional) name of the alias
         """
-        if not alias:
+        if alias is None:
             alias = application
-        self._snap("alias", [f"{self.name}.{application}", alias])
+        _cmd = ["snap", "alias", f"{self.name}.{application}", alias]
+        try:
+            subprocess.check_output(_cmd, universal_newlines=True)
+        except CalledProcessError as e:
+            raise SnapError(
+                "Snap: {!r}; command {!r} failed with output = {!r}".format(
+                    self._name, _cmd, e.output
+                )
+            )
 
     def restart(
         self, services: Optional[List[str]] = None, reload: Optional[bool] = False
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 0
+LIBPATCH = 1
 
 
 # Regex to locate 7-bit C1 ANSI sequences
@@ -273,7 +273,10 @@ class Snap(object):
           SnapError if there is a problem encountered
         """
         optargs = optargs or []
-        _cmd = ["snap", command, self._name, *optargs]
+        if command == "alias":
+            _cmd = ["snap", command, *optargs]
+        else:
+            _cmd = ["snap", command, self._name, *optargs]
         try:
             return subprocess.check_output(_cmd, universal_newlines=True)
         except CalledProcessError as e:
@@ -408,6 +411,17 @@ class Snap(object):
     def unhold(self) -> None:
         """Remove the refresh hold of a snap."""
         self._snap("refresh", ["--unhold"])
+
+    def alias(self, application: str, alias: str = None) -> None:
+        """Create an alias for a given application.
+
+        Args:
+            application: application to get an alias.
+            alias: (optional) name of the alias
+        """
+        if not alias:
+            alias = application
+        self._snap("alias", [f"{self.name}.{application}", alias])
 
     def restart(
         self, services: Optional[List[str]] = None, reload: Optional[bool] = False
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -273,13 +273,13 @@ class Snap(object):
           SnapError if there is a problem encountered
         """
         optargs = optargs or []
-        _cmd = ["snap", command, self._name, *optargs]
+        args = ["snap", command, self._name, *optargs]
         try:
-            return subprocess.check_output(_cmd, universal_newlines=True)
+            return subprocess.check_output(args, universal_newlines=True)
         except CalledProcessError as e:
             raise SnapError(
                 "Snap: {!r}; command {!r} failed with output = {!r}".format(
-                    self._name, _cmd, e.output
+                    self._name, args, e.output
                 )
             )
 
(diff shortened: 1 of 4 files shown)


lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -273,7 +273,10 @@ def _snap(self, command: str, optargs: Optional[Iterable[str]] = None) -> str:
           SnapError if there is a problem encountered
         """
         optargs = optargs or []
-        _cmd = ["snap", command, self._name, *optargs]
+        if command == "alias":
+            _cmd = ["snap", command, *optargs]
+        else:
+            _cmd = ["snap", command, self._name, *optargs]
```
dragomirp, line 279: Maybe I should create a separate method for this instead?

benhoyt: Yeah, I think so. This is an awkward special case. The `_snap` method is intended to run a snap-specific operation on a named snap. I would probably just put the `subprocess.run` call in `alias` itself.

benhoyt: There might also be a way to unify the various `subprocess.run` / `subprocess.check_output` calls so as to avoid repeating the `try/except` handling, but I don't know if it's worth it -- see what you think.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -409,6 +412,17 @@ def unhold(self) -> None:
         """Remove the refresh hold of a snap."""
         self._snap("refresh", ["--unhold"])
 
+    def alias(self, application: str, alias: str = None) -> None:
```
benhoyt, line 415: Seems like the type of `alias` should be `Optional[str]`.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -409,6 +412,17 @@ def unhold(self) -> None:
...
+        """Create an alias for a given application.
+
+        Args:
+            application: application to get an alias.
+            alias: (optional) name of the alias
+        """
+        if not alias:
```
benhoyt, line 422: I prefer the more explicit `if alias is None` in cases like this. Empty string is probably not valid here, but in general it might be, and it's good to distinguish between not provided (None) and set to empty string.

tests/unit/test_snap.py
```diff
@@ -661,3 +661,21 @@ def test_install_local_args(self, mock_subprocess):
                 universal_newlines=True,
             )
             mock_subprocess.reset_mock()
+
+    @patch("charms.operator_libs_linux.v2.snap.subprocess.check_output")
+    def test_alias(self, mock_subprocess):
+        mock_subprocess.return_value = 0
```
benhoyt, line 667: Shouldn't the return value of `check_output` be the output string? Probably `''` in this case.

lib/charms/operator_libs_linux/v2/snap.py
dragomirp: > There might also be a way to unify the various `subprocess.run` / `subprocess.check_output` calls so as to avoid repeating the `try/except` handling, but I don't know if it's worth it -- see what you think.

I don't think It's worth it for just a single call with try catch.

lib/charms/operator_libs_linux/v2/snap.py
dragomirp: Updated.

lib/charms/operator_libs_linux/v2/snap.py
dragomirp: Updated.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -409,6 +409,25 @@ def unhold(self) -> None:
         """Remove the refresh hold of a snap."""
         self._snap("refresh", ["--unhold"])
 
+    def alias(self, application: str, alias: Optional[str] = None) -> None:
+        """Create an alias for a given application.
+
+        Args:
+            application: application to get an alias.
+            alias: (optional) name of the alias
```
benhoyt, line 417: Let's add, "; if not provided, the application name is used."

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -409,6 +409,25 @@ def unhold(self) -> None:
...
+        """
+        if alias is None:
+            alias = application
+        _cmd = ["snap", "alias", f"{self.name}.{application}", alias]
```
benhoyt, line 421: No reason for this variable name to have an underscore prefix; `cmd` is fine. While we're here, in the [subprocess docs for `check_output`](https://docs.python.org/3/library/subprocess.html#subprocess.check_output), this is called `args`, so maybe that's a better name? (It's not just the command.)

I realize the other instances of this use `_cmd`, so maybe consistency is better than being different here. Alternatively, we could replace all the other instances of `_cmd` with `args` too. Up to you.

tests/integration/test_dnf.py
```diff
@@ -26,9 +26,9 @@ def test_query_dnf_and_package() -> None:
     assert package.name == "slurm-slurmd"
     assert package.arch == "x86_64"
     assert package.epoch is None
-    assert package.version == "22.05.6"
-    assert package.release == "3.el9"
-    assert package.full_version == "22.05.6-3.el9"
+    assert package.version == "22.05.9"
+    assert package.release == "1.el9"
+    assert package.full_version == "22.05.9-1.el9"
```
benhoyt, line 31: Two things:

1) Why are these numbers changing?
2) It's good to have unit tests, but I think we should also have an integration test for this. We already have [`integration/test_snap.py`](https://github.com/canonical/operator-libs-linux/blob/main/tests/integration/test_snap.py), so shouldn't be hard to add to that.

tests/unit/test_snap.py
```diff
@@ -661,3 +661,40 @@ def test_install_local_args(self, mock_subprocess):
...
+    @patch("charms.operator_libs_linux.v2.snap.subprocess.check_output")
+    def test_alias_raises_snap_error(self, mock_subprocess):
+        mock_subprocess.side_effect = CalledProcessError(
+            returncode=1, cmd=["snap", "alias", "foo.bar", "baz"]
+        )
+        foo = snap.Snap("foo", snap.SnapState.Latest, "stable", "1", "classic")
+        with self.assertRaises(snap.SnapError) as ctx:
+            foo.alias("bar", "baz")
+        self.assertEqual("<charms.operator_libs_linux.v2.snap.SnapError>", ctx.exception.name)
+        self.assertEqual(
```
benhoyt, line 692: I think testing the specifics of the error string is too much. It's not part of the API and users shouldn't be relying on it. I'd also probably drop the exception name test. We're already testing that it's an instance of `SnapError` -- that's the API contract, and seems to me all we should be testing here.

benhoyt (changes requested):
Looking good. A few minor comments, and a request for an integration test.

tests/unit/test_snap.py
dragomirp: Changed to only checking that the correct exception was raised.

tests/integration/test_dnf.py
dragomirp: >     1. Why are these numbers changing?

Most probably the package got updated in the images we use. The assertions fail without those changes. I can revert that, but then the integration suite will fail.

>     2. It's good to have unit tests, but I think we should also have an integration test for this. We already have [`integration/test_snap.py`](https://github.com/canonical/operator-libs-linux/blob/main/tests/integration/test_snap.py), so shouldn't be hard to add to that.

Added.

lib/charms/operator_libs_linux/v2/snap.py
dragomirp: I replaced them in the snap library. Best to have them consistent.

tests/integration/test_snap.py
```diff
@@ -237,3 +237,16 @@ def test_reset_hold_refresh():
...
+
+
+def test_alias():
+    cache = snap.SnapCache()
+    lxd = cache["lxd"]
+    lxd.alias("lxc", "testlxc")
+    result = check_output(["snap", "aliases"])
+    found_alias = False
+    for line in result.decode().split("\n")[1:]:
+        if ["lxd.lxc", "testlxc", "manual"] == [val for val in line.split(" ") if val]:
```
benhoyt, line 249: `line.split()` (without args) is [equivalent](https://docs.python.org/3/library/stdtypes.html#str.split) to `[val for val in line.split(" ") if val]`.

Also, the order of the equality test is a bit confusing ([Yoda conditions](https://en.wikipedia.org/wiki/Yoda_conditions) are uncommon in Go).

Could potentially simplify and remove the found flag by using `any(<generator expression>)`. Also, it would be good to include the output in the failure message to help debug any issues when they occur:

```python
found = any(line.split() == ["lxc.lxc", "testlxc", "manual"]
            for line in result.splitlines())
assert found, result
```

tests/integration/test_snap.py
```diff
@@ -237,3 +237,16 @@ def test_reset_hold_refresh():
     snap.hold_refresh(0)
     result = check_output(["snap", "refresh", "--time"])
     assert "hold: " not in result.decode()
+
+
+def test_alias():
```
benhoyt, line 242: Nice, thanks for this!

benhoyt (approved):
Approved, but just a couple minor suggestions to simplify the integration test code.

benhoyt:
I'll merge this tomorrow either way.

jnsgruk (approved):

tests/integration/test_snap.py
dragomirp: Thanks for the suggestion. Updated

jnsgruk:
Thanks @dragomirp - looks like you've addressed all of Ben's feedback so I've merged this! :)

PR #98, opened 2023-07-07
Add base logic for grub lib

rgildein:
- add load function for grub config (assumption no duplicate keys)
- add save function for grub config
- add is_container function to determine if running machine is container
- add GrubConfig object, which load config in lazy mode and provide function to make update (validate, merge and apply changes)
  - apply function with usage `grub-mkconfig` and `update-grub`
  - update function to wrap all functionality to update grub config
  - remove function if the charm is removed to remove charm configuration
 - add docstring with example

I tested this with simple ubuntu charm, where I tried to configure `GRUB_CMDLINE_LINUX_DEFAULT`. What I tested:
- install method with ValidationError (by manually changing `/etc/default/grub.d/95-juju-charm.cfg`
- update-status, where I check if current value is what charm expected
- remove method, when I removed the charm

```python
#!/usr/bin/env python3

import logging
from pathlib import Path
from subprocess import check_call, check_output
import charms.operator_libs_linux.v0.grub as grub

from ops.charm import CharmBase
from ops.main import main
from ops.model import ActiveStatus, BlockedStatus


log = logging.getLogger(__name__)


class UbuntuCharm(CharmBase):
    def __init__(self, *args):
        super().__init__(*args)
        ...
        ### testing of grub library
        self.framework.observe(self.on.install, self._on_install)
        self.framework.observe(self.on.update_status, self._on_update_status)
        self.framework.observe(self.on.remove, self._on_remove)
        self.grub = grub.GrubConfig(self.meta.name)
        log.info("found keys %s in grub config file", self.grub.keys())

   ...

    def _on_install(self, _):
        """Test configuring grub during install hook."""
        try:
            self.grub.update({"GRUB_CMDLINE_LINUX_DEFAULT": '"$GRUB_CMDLINE_LINUX_DEFAULT hugepagesz=1G"'})
        except grub.ValidationError as error:
            self.unit.status = BlockedStatus(f"[{error.key}] {error.message}")

    def _on_update_status(self, _):
        """Test checking grub configuration during update-status hook."""
        if self.grub["GRUB_CMDLINE_LINUX_DEFAULT"] != '"$GRUB_CMDLINE_LINUX_DEFAULT hugepagesz=1G"':
            self.unit.status = BlockedStatus("wrong grub configuration")

    def _on_remove(self, _):
        """Test cleaning grub configuration during remove hook."""
        self.grub.remove()


if __name__ == "__main__":
    main(UbuntuCharm)
```

### DIFF
--- a/tests/unit/test_grub.py
+++ b/tests/unit/test_grub.py
@@ -76,13 +76,13 @@ class BaseTestGrubLib(unittest.TestCase):
 
 class TestGrubUtils(BaseTestGrubLib):
     def test_split_config_line(self):
-        """Tets splitting single line."""
+        """Test splitting single line."""
         key, value = grub._split_config_line('test="1234"')
         assert key == "test"
         assert value == "1234"
 
     def test_split_config_line_failed(self):
-        """Tets splitting single line."""
+        """Test splitting single line."""
         with self.assertRaises(ValueError):
             grub._split_config_line('test="1234" "5678"')
 
@@ -210,19 +210,19 @@ class TestGrubConfig(BaseTestGrubLib):
         self.load_config.assert_not_called()
 
     def test__contains__(self):
-        """Tets config __contains__ function."""
+        """Test config __contains__ function."""
         self.assertIn("GRUB_TIMEOUT", self.config)
 
     def test__len__(self):
-        """Tets config __len__ function."""
+        """Test config __len__ function."""
         self.assertEqual(len(self.config), 4)
 
     def test__iter__(self):
-        """Tets config __iter__ function."""
+        """Test config __iter__ function."""
         self.assertListEqual(list(self.config), list(EXP_GRUB_CONFIG.keys()))
 
     def test__getitem__(self):
-        """Tets config __getitem__ function."""
+        """Test config __getitem__ function."""
         for key, value in EXP_GRUB_CONFIG.items():
             self.assertEqual(self.config[key], value)
 


lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -1,23 +1,31 @@
...
+# limitations under the License.
 
-This is a placeholder docstring for this charm library. Docstrings are
-presented on Charmhub and updated whenever you push a new version of the
-library.
+"""Simple library for managing Linux kernel configuration via grub.
 
-Complete documentation about creating and documenting libraries can be found
-in the SDK docs at https://juju.is/docs/sdk/libraries.
+NOTE: This library is not capable to read grub config file containes
```
dashmage, line 17: Is something missing at the end of the note?

`NOTE: This library is not capable of reading grub config files containing...`

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+
+
+class ApplyError(Exception):
+    """Exception if applying new config failed."""
+
+
+def _parse_config(data: str) -> Dict[str, str]:
+    """Parse config file lines.
+
+    This function is capable to update single key value if it's used with $ symbol.
```
dashmage, line 93: Second line of docstring might sound better like this:

`This function is capable of updating single key values using the "$" symbol`

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+        """Check if an update to /boot/grub/grub.cfg is available."""
+        try:
+            subprocess.check_call(f"grub-mkconfig -o {GRUB_CONFIG}", stderr=subprocess.STDOUT)
+            subprocess.check_call(["/usr/sbin/update-grub"], stderr=subprocess.STDOUT)
+        except subprocess.CalledProcessError as error:
+            logger.exception(error)
+            raise ApplyError("New config check failed.") from error
+
+    def _save_grub_configuration(self) -> None:
+        """Save current gru configuration."""
```
dashmage, line 196: typo: `gru` should be `grub`

esunar (changes requested):
We should add:
1. Note that this only handles simple configurations that starts with GRUB_
2. add GRUB_.* check for configuration keys

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+            if changed:
+                changed_keys.add(key)
+
+        return changed_keys
+
+    @property
+    def charm_name(self) -> str:
+        """Get charm name or use value obtained from JUJU_UNIT_NAME env."""
+        if self._charm_name is None:
+            self._charm_name, *_ = os.getenv("JUJU_UNIT_NAME", "unknown").split("/")
```
benhoyt, line 241: To me it seems odd that this is so closely tied to the charm name, particularly the use of environment variables. What's the need for having the charm name in this class? And if needed, I think it'd be preferable to require it to be passed in, rather than parsing environment variables in a library.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+    return False
+
+
+class GrubConfig(Mapping[str, str]):
+    """Grub config object.
+
+    This object will load current configuration option for grub and provide option
+    to update it with simple validation.
+
+    There is only one public function `update`, which should handle everything. It will
```
benhoyt, line 150: There's also `remove`, by the looks. Might be better not to mention the public methods here to avoid this getting out of date. Perhaps define the public methods at the top (after `__init__`) so they're more discoverable?

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+    changed keys to help charm check which keys where changed.
+    """
+
+    _lazy_data: Optional[Dict[str, str]] = None
+
+    def __init__(self, charm_name: Optional[str] = None) -> None:
+        """Initialize the grub config."""
+        self._charm_name = charm_name
+
+    def __contains__(self, key: str) -> bool:
```
benhoyt, line 161: Would it be simpler to just have `GrubConfig` subclass `dict` directly?

benhoyt: That said, I often find these "smart objects" cause more confusion than they bring value. It's kind of a dict, but not quite. Worth considering a dumber/simpler API that deals in plain dicts, perhaps top-level functions like `read_config() -> dict` and `update_config(dict)`?

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+    logger.info("grub config file %s was saved", path)
+
+
+def is_container() -> bool:
+    """Help function to see if local machine is container."""
+    # TODO: implement it
+    return False
+
+
+class GrubConfig(Mapping[str, str]):
```
benhoyt, line 144: I think we should avoid repeating the name of the module/lib in the class name. Especially for short import names, it's usually better to encourage `module.Name` use. So this would be just `Config`, and code would use it as `import grub ... grub.Config()`. See also https://benhoyt.com/writings/python-api-file-structure/

benhoyt (commented):
Looks like good code. Added some comments to consider.

lib/charms/operator_libs_linux/v0/grub.py
rgildein: This was part of not finished docstring. I update it, so you can check it now. Thanks

lib/charms/operator_libs_linux/v0/grub.py
rgildein: Can you check it again, I add full docstring with example.

rgildein:
> We should add:
> 
>     1. Note that this only handles simple configurations that starts with GRUB_
> 
>     2. add GRUB_.* check for configuration keys

I do not think that this is necessary, since this lib will be used by charm developer and not end user. If charm developer want, he/she can open the config file and do anything, that's why I do not see any benefits to add such validation.

Also without such validation we can do 
```json
{"TEST" : "1", "GRUB_A": "$TEST", "GRUB_B": "$TEST"}
```

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -1,23 +1,31 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
 
-This is a placeholder docstring for this charm library. Docstrings are
-presented on Charmhub and updated whenever you push a new version of the
-library.
+"""Simple library for managing Linux kernel configuration via grub.
```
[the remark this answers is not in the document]
rgildein: You mean that I should use `GRUB` instead of `grub`?

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
 # to 0 if you are raising the major API version
 LIBPATCH = 1
 
-# TODO: add your code here! Happy coding!
+LIB_CONFIG_DIRECTORY = Path("/var/lib/charm-grub/")
+GRUB_CONFIG = Path("/etc/default/grub.d/") / "95-juju-charm.cfg"
```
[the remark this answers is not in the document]
rgildein: I was using directory / path, than I removed the directory and now it's back as `GRUB_DIRECTORY`.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+
+# TODO: add better description to lib
+
+
+class ValidationError(ValueError):
+    """Exception representing value validation error."""
+
+    def __init__(self, key: str, message: str) -> None:
+        self._key = key
+        self._message = message
```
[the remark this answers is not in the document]
rgildein: Thanks that is good point, I'll do it.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +37,259 @@
...
+
+
+def _load_config(path: Path) -> Dict[str, str]:
+    """Load config file from /etc/default/grub.d/ directory."""
+    if not path.exists():
+        raise FileNotFoundError("grub config file %s was not found", path)
+
+    with open(path, "r", encoding="UTF-8") as file:
+        data = file.read()
+        config = _parse_config(data)
```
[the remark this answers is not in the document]
rgildein: Good point I will do it. It happens if you start with `_parse_config` function and than add `_load_config` :1234:

lib/charms/operator_libs_linux/v0/grub.py
rgildein: To be able to store only charm config file at `/etc/default/grub.d/90-juju-<charm-name>`. Using the env variable is only there so the charm developer does not have to think about it, we do not want to use `ops` here. Do you think that it's not needed? I can remove it.

lib/charms/operator_libs_linux/v0/grub.py
rgildein: You are ring, I forgot that I add `remove` and `apply`. I'll update docstring.
I'm used to define all public on bottom (my order magic objects (__init__, __eq__, ...), private properties, private functions, public properties and public method), but I'm ok to move it to top if that required.

lib/charms/operator_libs_linux/v0/grub.py
rgildein: It's not subclass of dict, only due to usage of lazy loading. If you check example in lib docstring on in this PR.
Basically the idea is to define `self.grub = GrubConfig()` in charm `__init__`, so it can be access in any function. Some charm may want to use `apply` in action instead of install, some maybe will do some configuration via config_changed.

lib/charms/operator_libs_linux/v0/grub.py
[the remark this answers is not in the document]
benhoyt: Yeah, GRUB, but only when we're referring to the project (per that doc, it's an acronym for GRand Unified Bootloader). But of course the module name will still be `grub`.

lib/charms/operator_libs_linux/v0/grub.py
benhoyt: Yeah, I would prefer charm libs don't parse environment variables directly, but that the charm name is passed in explicitly to the initialiser.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +70,288 @@
...
+# Original files:
+{configs}
+#
+# If you change this file, run 'update-grub' afterwards to update
+# /boot/grub/grub.cfg.
+# For full documentation of the options in this file, see:
+#   info -f grub -n 'Simple configuration'
+"""
+
+# TODO: use shlex
```
benhoyt, line 93: IMO this should block merge, otherwise it'll be very easy to get config files with invalid format.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +70,288 @@
...
+
+
+class ValidationError(ValueError):
+    """Exception representing value validation error."""
+
+    def __init__(self, key: str, message: str) -> None:
+        self.key = key
+        self.message = message
+        super().__init__(message)
```
benhoyt, line 102: Normally you'd call super first (eg: in case it sets `.message` and you want to overwrite it). Doesn't matter in this specific case, but it's not good practice.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +70,288 @@
...
+
+    context = [f"{key}={value}" for key, value in config.items()]
+    with open(path, "w", encoding="UTF-8") as file:
+        file.writelines([header, *context])
+
+    logger.info("GRUB config file %s was saved", path)
+
+
+def check_update_grub() -> bool:
+    """Check if an update to /boot/grub/grub.cfg is available."""
```
benhoyt, line 161: For boolean-returning functions, I like the word "Report whether an update..." as that indicates what the return value is. Similar for `is_container` below, "Report whether the local machine is a container."""

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +70,288 @@
...
+            ["/usr/bin/systemd-detect-virt", "--container"], stderr=subprocess.STDOUT
+        ).decode()
+        logger.debug("detect virt type %s", output)
+        return True
+    except subprocess.CalledProcessError:
+        return False
+
+
+class Config(Mapping[str, str]):
+    """Grub config object.
```
benhoyt, line 188:
```suggestion
    """Manages GRUB configuration.
```

lib/charms/operator_libs_linux/v0/grub.py
rgildein: Ok, I'm going to remove it.

lib/charms/operator_libs_linux/v0/grub.py
rgildein: You are right, I forgot to implement and I add only TODO for it. Thanks

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -1,24 +1,66 @@
...
+but these configurations will only have informational value as all configurations will be merged
+to "/etc/default/grub.d/95-juju-charm.cfg".
 
-Bear in mind that new revisions of the different major API versions (v0, v1,
-v2 etc) are maintained independently.  You can continue to update v0 and v1
-after you have pushed v3.
+Example of use:
 
-Markdown is supported, following the CommonMark specification.
+```python
```
benhoyt, line 27: Should this have an ending ``` before the end of the docstring?

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +71,300 @@
...
+        line = line.strip()
+        if not line or line.startswith("#"):
+            logger.debug("skipping line `%s`", line)
+            continue
+
+        key, value = line.split("=", 1)
+        if key in config:
+            logger.warning("key %s is duplicated in config", key)
+
+        config[key] = shlex.quote(value)
```
benhoyt, line 129: Should `_parse_config` be *un*quoting the value? Perhaps using `shlex.split`?

tests/unit/test_grub.py
```diff
@@ -0,0 +1,488 @@
...
+            GRUB_CONFIG_EXAMPLE + 'GRUB_CMDLINE_LINUX_DEFAULT="$GRUB_CMDLINE_LINUX_DEFAULT pti=on"'
+        )
+        stream = io.StringIO(raw_config)
+        result = grub._parse_config(stream)
+
+        self.logger.warning.assert_called_once_with(
+            "key %s is duplicated in config", "GRUB_CMDLINE_LINUX_DEFAULT"
+        )
+        self.assertEqual(
+            result["GRUB_CMDLINE_LINUX_DEFAULT"], "'\"$GRUB_CMDLINE_LINUX_DEFAULT pti=on\"'"
```
benhoyt, line 96: Yeah, per the comment above about quoting -- it's double-quoting things, when it should be unquoting the value. I would expect this:

```python
 self.assertEqual(
            result["GRUB_CMDLINE_LINUX_DEFAULT"], "$GRUB_CMDLINE_LINUX_DEFAULT pti=on",
)
```

tests/unit/test_grub.py
```diff
@@ -0,0 +1,488 @@
...
+
+    @mock.patch.object(grub, "_parse_config")
+    def test_load_config(self, mock_parse_config):
+        """Test load config from file."""
+        exp_config = {"test": "valid"}
+        mock_parse_config.return_value = exp_config
+        path = self.tmp_dir / "test_load_config"
+        path.touch()  # create file
+
+        with mock.patch.object(grub, "open", mock.mock_open()) as mock_open:
```
benhoyt, line 114: I understand wanting to mock the subprocess calls, but why do we need to mock the open calls? Especially seeing the test is creating temp directories -- can't we do real open/read on real files in the tests? (Local FS operations are fast and reliable.) Or am I missing something here.

benhoyt (changes requested):

tests/unit/test_grub.py
rgildein: There is no point to actually reading it, since `_parse_config` is mocked and I want to check that open was called correctly.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -1,24 +1,67 @@
...
+        self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.update_status, self._on_update_status)
+        self.framework.observe(self.on.remove, self._on_remove)
+        self.grub = grub.GrubConfig(self.meta.name)
+        log.debug("found keys %s in GRUB config file", self.grub.keys())
+
+    def _on_install(self, _):
+        try:
+            self.grub.update(
+                {"GRUB_CMDLINE_LINUX_DEFAULT": '"$GRUB_CMDLINE_LINUX_DEFAULT hugepagesz=1G"'}
```
benhoyt, line 40: Presumably this example shouldn't have the quoting, given that we're passing the key-value strings? Similar below in `_on_update_status`.

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -29,4 +72,300 @@
...
+        line = line.strip()
+        if not line or line.startswith("#"):
+            logger.debug("skipping line `%s`", line)
+            continue
+
+        key, value = line.split("=", 1)
+        if key in config:
+            logger.warning("key %s is duplicated in config", key)
+
+        config[key], *_ = shlex.split(value)
```
benhoyt, line 130: It [sounds like](https://www.gnu.org/software/grub/manual/grub/html_node/Simple-configuration.html#Simple-configuration) `key=foo bar` is not valid? It would have to be quoted in that case, right.

I think we should issue a warning if we get things like `key=foo bar` or `key="foo bar" "baz"`, that is, more than one value. Or maybe it should even be an exception?

```python
    values = shlex.split(value)
    if len(values) > 1:
        logger.warning("key %s has more than one value", key)
    config[key] = values[0]
```

benhoyt: Might also be worth sending a bunch of existing GRUB config file examples through this and see that it parses them as expected.

benhoyt: I'm no expert on GRUB, but are we also going to run into trouble if the GRUB config uses scripting features? https://www.gnu.org/software/grub/manual/grub/html_node/Shell_002dlike-scripting.html

benhoyt (approved):
Left a couple of minor comments, but it looks reasonably good to me. I'm not an expert in GRUB, so might pay to get someone to review that is, but maybe one of the other reviewers that's already looked at it has done that?

lib/charms/operator_libs_linux/v0/grub.py
rgildein: You cannot do `GRUB_TERMINAL_INPUT=console serial` or `GRUB_TERMINAL_INPUT="console" "serial"`, because `grub-mkconfig` will failed with following error:
```bash
$ sudo grub-mkconfig -o /tmp/test
Sourcing file `/etc/default/grub'
Sourcing file `/etc/default/grub.d/50-cloudimg-settings.cfg'
Sourcing file `/etc/default/grub.d/95-juju-charm.cfg'
/usr/sbin/grub-mkconfig: 14: /etc/default/grub.d/95-juju-charm.cfg: serial: not found
```
That's why I completely ignore other values, since there should never be such situation. At least the lib will never created it, only some user can do it manually. Do you think we should check such situation? If yes I think we should raise and exception.

> I'm no expert on GRUB, but are we also going to run into trouble if the GRUB config uses scripting features? 

Me neither, but I think that this lib do not need to support it. Maybe someone will come and request it as feature, but for now it's ok to not supported. What do you say?

lib/charms/operator_libs_linux/v0/grub.py
rgildein: I'll updated it. Thanks

lib/charms/operator_libs_linux/v0/grub.py
benhoyt: Yeah, I think we should raise an exception if `shlex.split()` returns more than one value then.

Re scripting -- sounds good to ignore that and not support it for now.

benhoyt (approved):
LGTM (one nit comment). Do you think you're ready to merge this?

rgildein:
> LGTM (one nit comment). Do you think you're ready to merge this?

I think it's ready. Is there anyone else who should do review?

PR #99, opened 2023-07-12
Sysctl init

zmraul:
PoC integration with Kafka: https://github.com/canonical/kafka-operator/pull/118
Addresses [DPE-2111](https://warthogs.atlassian.net/browse/DPE-2111)

This draft will be updated with feedback from #98 and Kafka PR before setting it to ready to review.

[DPE-2111]: https://warthogs.atlassian.net/browse/DPE-2111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/charms/operator_libs_linux/v0/sysctl.py
+++ b/lib/charms/operator_libs_linux/v0/sysctl.py
@@ -12,15 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Handler for the sysctl config
-"""
+"""Handler for the sysctl config."""
 
+import glob
 import logging
 import os
 import re
-import glob
-from dataclasses import dataclass
+from pathlib import Path
 from subprocess import STDOUT, CalledProcessError, check_output
+from typing import Dict, Mapping, Optional
 
 logger = logging.getLogger(__name__)
 
@@ -35,39 +35,12 @@ LIBAPI = 0
 LIBPATCH = 2
 
 
-SYSCTL_DIRECTORY = "/home/raul/workspace/operator-libs-linux"
-SYSCTL_FILENAME = f"{SYSCTL_DIRECTORY}/95-juju-sysctl.conf"
+SYSCTL_DIRECTORY = Path("/etc/sysctl.d")
+SYSCTL_FILENAME = Path(f"{SYSCTL_DIRECTORY}/95-juju-sysctl.conf")
 SYSCTL_HEADER = f"""# This config file was produced by sysctl lib v{LIBAPI}.{LIBPATCH}
 #
 # This file represents the output of the sysctl lib, which can combine multiple
-# configurations into a single file like this with the help of validation rules. Such
-# rules are used to automatically resolve simple conflicts between two or more charms
-# on one host and is defined as a comment after the configuration option, see example
-# below.
-#
-# Description of validation rules to check if values are valid.
-#   - "range(a,b)" - all numbers between 'a' (included) and 'b' (excluded)
-#   - "a|b|c" - choices 'a', 'b' and 'c'
-#   - "*" - any value
-#   - "" or no comment - only current value, same as "<current value>"
-#   - "disable" - This value will be ignored in any validation and should only be used by
-#                 the charm operator manually or via a charm to override any system
-#                 configuration.
-# Examples:
-# # any value in [10, 50] is valid value
-# vm.swappiness=10  # range(10,51)
-#
-# # 60 and 80 are valid values
-# vm.dirty_ratio = 80  # 60|80
-#
-# # any value is valid
-# vm.dirty_background_ratio = 3  # *
-#
-# # only value 1 is valid value
-# net.ipv6.conf.all.use_tempaddr = 1
-#
-# # validation is disabled
-# net.ipv6.conf.default.use_tempaddr = 0  # disable
+# configurations into a single file like.
 """
 
 
@@ -88,6 +61,7 @@ class Error(Exception):
         """Return the message passed as an argument."""
         return self.args[0]
 
+
 class SysctlError(Error):
     """Raised when there's an error running sysctl command."""
 
@@ -97,46 +71,48 @@ class SysctlPermissionError(Error):
 
 
 class ValidationError(Error):
-    """Raised when there's an error in the validation process."""
+    """Exception representing value validation error."""
 
 
-@dataclass()
-class ConfigOption:
-    """Definition of a config option.
-    
-    NOTE: this class can be extended to handle the rule tied to each option.
-    """
-    name: str
-    value: int
-
-    @property
-    def string_format(self) -> str:
-        return f"{self.name}={self.value}"
-
-
-class SysctlConfig:
+class SysctlConfig(Mapping[str, int]):
     """Represents the state of the config that a charm wants to enforce."""
 
-    def __init__(self, config_params: dict, app_name: str) -> None:
-        self.config_params = config_params
-        self.app_name = app_name
+    def __init__(self, name: Optional[str] = None) -> None:
+        self.name = name
+        self._data = self._load_data()
+
+    def __contains__(self, key: str) -> bool:
+        """Check if key is in config."""
+        return key in self._data
+
+    def __len__(self):
+        """Get size of config."""
+        return len(self._data)
+
+    def __iter__(self):
+        """Iterate over config."""
+        return iter(self._data)
+
+    def __getitem__(self, key: str) -> int:
+        """Get value for key form config."""
+        return self._data[key]
 
     @property
-    def config_params(self) -> list[ConfigOption]:
-        """Config options passed to the lib."""
-        return self._config_params
+    def name(self) -> str:
+        """Name used to create the lib file."""
+        return self._name
 
-    @config_params.setter
-    def config_params(self, value: dict):
-        result = []
-        for k, v in value.items():
-            result += [ConfigOption(k, v["value"])]
-        self._config_params = result
+    @name.setter
+    def name(self, value: str):
+        if value is None:
+            self._name, *_ = os.getenv("JUJU_UNIT_NAME", "unknown").split("/")
+        else:
+            self._name = value
 
     @property
-    def charm_filepath(self) -> str:
+    def charm_filepath(self) -> Path:
         """Name for resulting charm config file."""
-        return f"{SYSCTL_DIRECTORY}/90-juju-{self.app_name}"
+        return Path(f"{SYSCTL_DIRECTORY}/90-juju-{self.name}")
 
     @property
     def charm_config_exists(self) -> bool:
@@ -148,52 +124,61 @@ class SysctlConfig:
         """Return whether a merged config file exists."""
         return os.path.exists(SYSCTL_FILENAME)
 
-    @property
-    def merged_config(self) -> list[ConfigOption] | None:
-        """Return applied internal config."""
-        if not self.merged_config_exists:
-            return None
+    def update(self, config: Dict[str, dict]) -> None:
+        """Update sysctl config options with a desired set of config params."""
+        self._parse_config(config)
 
-        with open(SYSCTL_FILENAME, "r") as f:
-            return [ConfigOption(param.strip(), value.strip())
-                    for line in f.read().splitlines()
-                    if line and not line.startswith('#')
-                    for param, value in [line.split('=')]]
-
-    def update(self) -> None:
-        """Update sysctl config options."""
+        # NOTE: case where own charm calls update() more than once. Remove first so
+        # we don't get validation errors.
         if self.charm_config_exists:
-            return
-        self._create_charm_file()
+            self.remove()
 
-        if not self.validate():
-            raise ValidationError()
+        conflict = self._validate()
+        if conflict:
+            msg = f"Validation error for keys: {conflict}"
+            raise ValidationError(msg)
 
         snapshot = self._create_snapshot()
-        print(f"CREATED SNAPSHOT: {snapshot}")
+        logger.debug(f"Created snapshot for keys: {snapshot}")
         try:
             self._apply()
         except SysctlPermissionError:
             self._restore_snapshot(snapshot)
             raise
-        except ValidationError:
+        except SysctlError:
             raise
+
+        self._create_charm_file()
         self._merge()
 
-    def validate(self) -> bool:
+    def remove(self) -> None:
+        """Remove config for charm."""
+        self.charm_filepath.unlink(missing_ok=True)
+        logger.info("charm config file %s was removed", self.charm_filepath)
+        self._merge()
+
+    def _validate(self) -> list[str]:
         """Validate the desired config params against merged ones."""
-        return True
+        common_keys = set(self._data.keys()) & set(self._desired_config.keys())
+        confict_keys = []
+        for key in common_keys:
+            if self._data[key] != self._desired_config[key]:
+                msg = f"Values for key '{key}' are different: {self._data[key]} != {self._desired_config[key]}"
+                logger.warning(msg)
+                confict_keys.append(key)
+
+        return confict_keys
 
     def _create_charm_file(self) -> None:
         """Write the charm file."""
-        charm_params = [f"{param.string_format}\n" for param in self.config_params]
+        charm_params = [f"{key}={value}\n" for key, value in self._desired_config.items()]
         with open(self.charm_filepath, "w") as f:
             f.writelines(charm_params)
 
     def _merge(self) -> None:
         """Create the merged sysctl file."""
         # get all files that start by 90-juju-
-        charm_files = [f for f in glob.glob(f"{SYSCTL_DIRECTORY}/90-juju-*")]
+        charm_files = list(glob.glob(f"{SYSCTL_DIRECTORY}/90-juju-*"))
         data = [SYSCTL_HEADER]
         for path in charm_files:
             with open(path, "r") as f:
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+            config: dictionary with keys to update:
+        ```
+        {"vm.swappiness": {"value": 10}, ...}
+        ```
+        """
+        self._parse_config(config)
+
+        # NOTE: case where own charm calls update() more than once.
+        if self.charm_filepath.exists():
+            self._merge(add_own_charm=False)
```
rgildein, line 173: I think we need to provide better description in doc-string or change logic here, because this function is called `update` and it's not following real update pattern. I'll try to explain it with example.
Calling `update` twice as shown bellow, does not ensure that `95-juju-sysctl.conf` will have `vm.max_map_count` configured. It will be applied, but there will be no information and after reboot it will be lost.
```python
self.sysctl.update({"vm.max_map_count": 262144})
self.sysctl.update({"vm.swappiness": 50})
```
From this situation I would expect that `95-juju-sysctl.conf` to have both `vm.max_map_count=262144` and `vm.swappiness=50`.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+            self._apply()
+        except SysctlPermissionError:
+            self._restore_snapshot(snapshot)
+            raise
+
+        self._create_charm_file()
+        self._merge()
+
+    def remove(self) -> None:
+        """Remove config for charm."""
```
rgildein, line 191: I would prefer to add note about not applying anything (since we are not doing any change in sysctl configuration, not reverting any value) with this remove, so charm developer will not expect that after calling `sysctl.remove()`.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+        logger.debug("Failed values: %s", failed_values)
+
+        if failed_values:
+            msg = f"Unable to set params: {[f.group(1) for f in failed_values]}"
+            logger.error(msg)
+            raise SysctlPermissionError(msg)
+
+    def _create_snapshot(self) -> Dict[str, str]:
+        """Create a snaphot of config options that are going to be set."""
+        return {key: self._sysctl([key, "-n"])[0] for key in self._desired_config.keys()}
```
rgildein, line 254: Like this we will call subprocess to many times and this can easily took 10+ seconds. Can we call it only once or load those values from `95-juju-sysctl.cong`?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+        if not SYSCTL_FILENAME.exists():
+            return config
+
+        with open(SYSCTL_FILENAME, "r") as f:
+            for line in f:
+                config.update(self._parse_line(line))
+
+        return config
+
+    def _parse_line(self, line: str) -> Dict[str, str]:
```
rgildein, line 291: To me it would be much cleaner if this function returns `Tuple[str, str]` and not dictionary with single key and value.

tests/unit/test_sysctl.py
```diff
@@ -0,0 +1,332 @@
...
+    # Tests on 'update()'
+    elif args[0] == ["sysctl", "vm.max_map_count", "-n"]:
+        return "25000"
+    elif args[0] == ["sysctl", "vm.max_map_count=25500"]:
+        return "25500"
+
+
+class TestSysctlConfig(unittest.TestCase):
+    def setUp(self) -> None:
+        self.desired_values = {"vm.swappiness": {"value": 0}}
```
rgildein, line 61: this variable is never used, can we remove it

tests/unit/test_sysctl.py
```diff
@@ -0,0 +1,332 @@
...
+
+    # Tests on 'update()'
+    elif args[0] == ["sysctl", "vm.max_map_count", "-n"]:
+        return "25000"
+    elif args[0] == ["sysctl", "vm.max_map_count=25500"]:
+        return "25500"
+
+
+class TestSysctlConfig(unittest.TestCase):
+    def setUp(self) -> None:
```
rgildein, line 60: I really think that we should mocked `SYSCTL_DIRECTORY` (probably also `SYSCTL_FILENAME`) to ensure that this tests will not create such file or change it. Right now I did not spotted any place where we are not mocking load or write config, but this can easily happen in future and we should avoid touching system configuration in unit tests.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+    def _sysctl(self, cmd: List[str]) -> List[str]:
+        """Execute a sysctl command."""
+        cmd = ["sysctl"] + cmd
+        logger.debug("Executing sysctl command: %s", cmd)
+        try:
+            return check_output(cmd, stderr=STDOUT, universal_newlines=True).splitlines()
+        except CalledProcessError as e:
+            msg = f"Error executing '{cmd}': {e.stdout}"
+            logger.error(msg)
+            raise SysctlError(msg)
```
rgildein, line 270:
```suggestion
            raise SysctlError(msg) from error
```

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+    def message(self):
+        """Return the message passed as an argument."""
+        return self.args[0]
+
+
+class SysctlError(Error):
+    """Raised when there's an error running sysctl command."""
+
+
+class SysctlPermissionError(Error):
```
rgildein, line 124: not-blocking: Can we stay on same page here and use `ApplyError` as I did in grub lib? To me it's more obvious than some permission error.

rgildein (commented):
I think that we could improve some small thinks, but it's overall it looks good to me.
I will also suggest to add more logging, since we are not doing much of that.

tests/unit/test_sysctl.py
```diff
@@ -0,0 +1,332 @@
...
+        self.desired_values = {"vm.swappiness": {"value": 0}}
+        self.loaded_values = {"vm.swappiness": "60", "vm.max_map_count": "25500"}
+
+    @patch("pathlib.Path.exists")
+    @patch("pathlib.Path.glob")
+    @patch("builtins.open", new_callable=mock_open, read_data="vm.max_map_count=25500\n")
+    @patch("charms.operator_libs_linux.v0.sysctl.check_output")
+    @patch("charms.operator_libs_linux.v0.sysctl.Config._create_charm_file")
+    @patch("charms.operator_libs_linux.v0.sysctl.Config._load_data")
```
benhoyt, line 69: Yeah, this is a lot of mocking... I think we'd be better to just change the directory and use real file operations. We're testing a lot more that way. We can have one test with mocked file operations if you want, to ensure that we're actually writing to the right path if the path isn't overridden.

benhoyt (commented):
I haven't looked at the test thoroughly yet, but I added a few initial comments on the code.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+        self.charm_filepath.unlink(missing_ok=True)
+        logger.info("Charm config file %s was removed", self.charm_filepath)
+        self._merge()
+
+    def _validate(self) -> List[str]:
+        """Validate the desired config params against merged ones."""
+        common_keys = set(self._data.keys()) & set(self._desired_config.keys())
+        conflict_keys = []
+        for key in common_keys:
+            if self._data[key] != self._desired_config[key]:
```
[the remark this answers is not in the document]
zmraul: The validation is run against keys that are by other charms using the lib. This is to avoid conflicts when two or more charms that set keys are deployed on the same machine.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+        return config
+
+    def _parse_line(self, line: str) -> Dict[str, str]:
+        """Parse a line from juju-sysctl.conf file."""
+        if line.startswith("#") or line == "\n":
+            return {}
 
-# TODO: add your code here! Happy coding!
+        param, value = line.split("=")
+        return {param.strip(): value.strip()}
```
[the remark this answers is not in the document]
zmraul: This was done initially on _load as you point out, but it looked something like:
```pyhton
{
    param.strip(): value.strip()
    for line in f.read().splitlines()
    if line and not line.startswith('#')
    for param, value in [line.split('=')]
}
```
which is quite a mouthful :/

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Yes, maybe `update` is not the correct term here. It's more a `configure` scenario, where we change everything, not adding.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+
+        # Reload data with newly created file.
+        self._data = self._load_data()
+
+    def _apply(self) -> None:
+        """Apply values to machine."""
+        cmd = [f"{key}={value}" for key, value in self._desired_config.items()]
+        result = self._sysctl(cmd)
+        expr = re.compile(r"^sysctl: permission denied on key \"([a-z_\.]+)\", ignoring$")
+        failed_values = [expr.match(line) for line in result if expr.match(line)]
```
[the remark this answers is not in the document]
zmraul: Good to know, I'll move this to the top

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+        """Return a string representation of the model plus class."""
+        return "<{}.{}>".format(type(self).__module__, type(self).__name__)
+
+    @property
+    def message(self):
+        """Return the message passed as an argument."""
+        return self.args[0]
+
+
+class SysctlError(Error):
```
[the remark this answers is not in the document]
zmraul: I think `CommandError` might be best here

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
+# configurations into a single file like.
+"""
+
+
+class Error(Exception):
+    """Base class of most errors raised by this library."""
+
+    def __repr__(self):
+        """Represent the Error."""
+        return "<{}.{} {}>".format(type(self).__module__, type(self).__name__, self.args)
```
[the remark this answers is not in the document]
zmraul: I'm guilty of copying the exception format from the other modules in here :) I'll change it to remove unused parts. Thanks for the explanation!

tests/unit/test_sysctl.py
zmraul: It is a lot, but I was trying to avoid creating files here. This happens more on the `update` call which is the one that holds most of the logic. Would adding integration tests (and leaving unit tests like this) cover your concern here?

lib/charms/operator_libs_linux/v0/sysctl.py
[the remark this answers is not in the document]
zmraul: nvm, I merged `_parse_line` into load function now.

lib/charms/operator_libs_linux/v0/sysctl.py
benhoyt: Yes, if the behaviour here is to write an entire config, I agree with @rgildein -- this name is confusing. It's kind of like `dict.update`, but does something very different. What about renaming it `configure`, as you suggest?

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,203 @@
...
+                )
+                conflict_keys.append(key)
+
+        return conflict_keys
+
+    def _create_charm_file(self) -> None:
+        """Write the charm file."""
+        charm_params = [f"{key}={value}\n" for key, value in self._desired_config.items()]
+        with open(self.charm_filepath, "w") as f:
+            f.writelines(charm_params)
```
benhoyt, line 215: It's a minor point for small config files like this, but it's usually better to write a single line at a time, avoiding the whole file being instantiated in memory at once. It's also a bit simpler. Something like this:

```python
with open(self.charm_filepath, "w") as f:
    for key, value in self._desired_config.items():
        f.write(f"{key}={value}\n")
```

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,203 @@
...
+            msg = f"Error executing '{cmd}': {e.stdout}"
+            logger.error(msg)
+            raise CommandError(msg)
+
+    def _parse_config(self, config: Dict[str, dict]) -> None:
+        """Parse a config passed to the lib."""
+        result = {}
+        for key, value in config.items():
+            result[key] = str(value["value"])
+        self._desired_config: Dict[str, str] = result
```
benhoyt, line 279: Nit: good use case for a dictionary comprehension:

```python
self._desired_config = {k: str(v) for k, v in config.items()}
```

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,211 @@
...
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
+
+CHARM_FILENAME_PREFIX = "90-juju-"
+SYSCTL_DIRECTORY = Path("/etc/sysctl.d")
+SYSCTL_FILENAME = Path(f"{SYSCTL_DIRECTORY}/95-juju-sysctl.conf")
+SYSCTL_HEADER = f"""# This config file was produced by sysctl lib v{LIBAPI}.{LIBPATCH}
+#
+# This file represents the output of the sysctl lib, which can combine multiple
```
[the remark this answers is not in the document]
zmraul: I added this as part of file creation, see this commit: https://github.com/canonical/operator-libs-linux/pull/99/commits/0cd98169b149cd4866dfb350a6844f7d44e9bc3b

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,209 @@
...
+        ]
+        logger.debug("Failed values: %s", failed_values)
+
+        if failed_values:
+            msg = f"Unable to set params: {[f.group(1) for f in failed_values]}"
+            logger.error(msg)
+            raise ApplyError(msg)
+
+    def _create_snapshot(self) -> Dict[str, str]:
+        """Create a snaphot of config options that are going to be set."""
```
benhoyt, line 253: Nit: snaphot -> snapshot

lib/charms/operator_libs_linux/v0/sysctl.py
zmraul: Added as part of the description

benhoyt (changes requested):
Thanks for the updates. The main thing I'd want to see changed is the signature of `update` to not use the inner `{"value": x}` one-key dict (and probably change the name).

I'd also much prefer the tests to avoid so much mocking of internals (and filesystem operations): if we could just mock the subprocess call and use real filesystem operations for the file handling, the tests would be more robust in the face of implementation changes, and we'd need fewer integration tests.

Let's also get some end-to-end / integration tests here. I think we only need a few, especially if we update the unit tests to use real filesystem operations.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,203 @@
...
+        """Name for resulting charm config file."""
+        return SYSCTL_DIRECTORY / f"{CHARM_FILENAME_PREFIX}{self.name}"
+
+    def update(self, config: Dict[str, dict]) -> None:
+        """Update sysctl config options with a desired set of config params.
+
+        Args:
+            config: dictionary with keys to update:
+        ```
+        {"vm.swappiness": {"value": 10}, ...}
```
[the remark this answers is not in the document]
benhoyt: I would suggest making a good API for what we have now then (that is, avoid the "value" dict), and then later, if we extend it, we could use something like `{"vm.swappiness": Rule(foo, bar)}`.

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,209 @@
...
+        data = [SYSCTL_HEADER]
+        paths = set(SYSCTL_DIRECTORY.glob(f"{CHARM_FILENAME_PREFIX}*"))
+        if not add_own_charm:
+            paths.discard(self.charm_filepath.as_posix())
+
+        for path in paths:
+            with open(path, "r") as f:
+                data += f.readlines()
+        with open(SYSCTL_FILENAME, "w") as f:
+            f.writelines(data)
```
rgildein, line 233: not-blocking: follow up from @benhoyt comment about creating charm file.
```suggestion
        with open(SYSCTL_FILENAME, "w") as f:
            f.write(SYSCTL_HEADER)
            for path in paths:
                with open(path, "r") as charm_file:
                    f.write(charm_file.readlines())
```

lib/charms/operator_libs_linux/v0/sysctl.py
```diff
@@ -27,6 +87,209 @@
...
+        config = {}
+        if not SYSCTL_FILENAME.exists():
+            return config
+
+        with open(SYSCTL_FILENAME, "r") as f:
+            for line in f:
+                if line.startswith(("#", ";")) or not line.strip() or "=" not in line:
+                    continue
+
+                key, _, value = line.partition("=")
```
rgildein, line 292: not-blocking: 
```suggestion
                key, value = line.split("=", 1)
```

tests/unit/test_sysctl.py
rgildein: I'm not against so many `@patch`, but I still think that we should change `SYSCTL_DIRECTORY` to some tmp path to be sure that we will not change anything in system sysctl config and maybe also mocked all subprocess calls so unit tests will not call `sysctl key=value`. I meant in `setUp` function.

rgildein (commented):
Two not-blocking comment + one about unit tests.

benhoyt (approved):
This looks a lot better to me now, thanks!

rgildein (approved):
LGTM

benhoyt:
All good to merge, or does anyone else need to review this? If not, I can merge.

zmraul:
> All good to merge, or does anyone else need to review this? If not, I can merge.

I think we are good to merge. Thanks!

PR #101, opened 2023-07-26
Fix error naming example

zmraul:
Fixes docstring to have correct exception naming.

### DIFF
--- a/tests/integration/test_sysctl.py
+++ b/tests/integration/test_sysctl.py
@@ -8,16 +8,6 @@ from subprocess import check_output
 
 from charms.operator_libs_linux.v0 import sysctl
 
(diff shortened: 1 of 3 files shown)


tests/integration/test_sysctl.py
```diff
@@ -8,7 +8,7 @@
 
 from charms.operator_libs_linux.v0 import sysctl
 
-EXPECTED_MERGED_RESULT = """# This config file was produced by sysctl lib v0.2
+EXPECTED_MERGED_RESULT = """# This config file was produced by sysctl lib v0.3
```
benhoyt, line 11: Can we use `v{LIBAPI}.{LIBPATCH}` here (and below), to avoid having to update this every update?

zmraul: Yes, that is much better.

benhoyt (approved):

PR #102, opened 2023-08-01
Increate patch version fro grub lib

rgildein:
Fix patch version, so the it's possible to fetch new version.

I forgot to increase patch version in #97.


benhoyt (approved):

benhoyt:
Thanks!

PR #104, opened 2023-08-04
Add functional tests for grub lib

rgildein:
- add functional tests for grub lib
  - the feature tested for grub is required to run on a **VM** instead of a **container**, so I changed the command in tox to use a VM with LXD
- add `_update_config` function to update charm config file instead of overwriting it
- fix writing lines in `_save_config` function
- raise ApplyError if `grub-mkconfig` failed, since it is failing if wrong configuration was provided


lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -161,13 +161,24 @@ def _save_config(path: Path, config: Dict[str, str], header: str = CONFIG_HEADER
     if path.exists():
         logger.debug("GRUB config %s already exist and it will overwritten", path)
 
-    context = [f"{key}={shlex.quote(value)}" for key, value in config.items()]
     with open(path, "w", encoding="UTF-8") as file:
-        file.writelines([header, *context])
+        file.write(f"{header}{os.linesep}")
```
benhoyt, line 165: There's no need to use `os.linesep` here, because we know we're on Linux / POSIX, so `\n` is simpler and clearer. In addition, as the [docs for `os.linesep`](https://docs.python.org/3/library/os.html#os.linesep) say, "Do not use os.linesep as a line terminator when writing files opened in text mode (the default); use a single '\n' instead, on all platforms."

```suggestion
        file.write(header + "\n")
```

Or maybe the header string could just include the `\n` in it?

lib/charms/operator_libs_linux/v0/grub.py
```diff
@@ -161,13 +161,24 @@ def _save_config(path: Path, config: Dict[str, str], header: str = CONFIG_HEADER
...
+        for key, value in config.items():
+            file.write(f"{key}={shlex.quote(value)}{os.linesep}")
```
benhoyt, line 167:
```suggestion
            file.write(f"{key}={shlex.quote(value)}\n")
```
But yeah, this new code structure is definitely clearer than what it was before.

tests/unit/test_grub.py
```diff
@@ -138,7 +156,12 @@ def test_save_config(self):
             grub._save_config(path, {"test": '"1234"'})
 
         mock_open.assert_called_once_with(path, "w", encoding="UTF-8")
-        mock_open.return_value.writelines.assert_called_once_with([mock.ANY, "test='\"1234\"'"])
+        mock_open.return_value.write.assert_has_calls(
+            [
+                mock.call(f"{grub.CONFIG_HEADER}{os.linesep}"),
+                mock.call(f"test='\"1234\"'{os.linesep}"),
```
benhoyt, line 162: Similar here: no need for `os.linesep`, just use `\n`.

tests/unit/test_grub.py
```diff
@@ -358,144 +394,233 @@ def test_remove_no_config(self, mock_save, mock_apply):
...
+        mock_update.assert_not_called()
 
     @mock.patch.object(grub.Config, "apply")
     @mock.patch.object(grub.Config, "_save_grub_configuration")
-    @mock.patch.object(
-        grub.Config, "_set_value", side_effect=grub.ValidationError("test", "test_message")
-    )
-    def test_update_validation_failure(self, _, mock_save, mock_apply):
+    @mock.patch.object(grub.Config, "_update")
+    @mock.patch.object(grub, "_update_config")
```
benhoyt, line 440: I still don't like all this mocking. It makes the tests more fragile as we're testing a lot more test internals. Ideally we could change these to use the real filesystem (in a temp directory) where possible, and only use mocking where necessary (for example, the `subprocess` update-grub calls). Local filesystem operations are fast and reliable, so don't need to be mocked.

benhoyt (changes requested):
Change itself looks reasonable. One small request (avoiding `os.linesep`) and one larger request (that we avoid mocking the filesystem in unit tests). However, it might be too much of a change in this PR to change how we're testing filesystem operations, so I'd be happy to keep that in mind for future work.

lib/charms/operator_libs_linux/v0/grub.py
rgildein: Actually header is providing `\n`, so it's not needed. Regarding `os.linesep`, I read the docs and I must say thanks for pointing this out. I'm always saying everyone to use it everywhere, it's it's not desired at all. I'll replace it with `\n`.

tests/unit/test_grub.py
rgildein: Hi @benhoyt, I agree, but at the same time the definition of unit tests is testing functionality of single function, because it's not good if I change one thing and it breaks half of unit tests. So for me it's really necessary to tests each function deeply with mocked everything else.
However at the beginning I mention that I agree with you and that's why I provide `TestSmokeConfig` where I mocked everything and `TestFullConfig` where I mocked only subprocess check_call. I think that it is good solution, because like this I test all scenarios for each function with mocking other functions and the general usage for `.remove` and `.update` function.
What do you think?

rgildein:
@benhoyt Can me moved CI to runner, which will be able to run VMs with lxd? It's mandatory to run integration tests for grub on non container machine, or potentially we can mocked that function.

tests/unit/test_grub.py
benhoyt: > I agree, but at the same time the definition of unit tests is testing functionality of single function, because it's not good if I change one thing and it breaks half of unit tests. So for me it's really necessary to tests each function deeply with mocked everything else.

"Unit testing" is a somewhat vague term, but just given the name, I don't think it necessarily means testing functions, but testing *units* (which could be modules, classes, methods). That said, I don't believe that debating the term is particularly helpful. I agree with you that we want to avoid "if I change one thing it breaks half of the tests".

But, as I see it, mocking internal helper functions is exactly how you'd get that problem. This test is mocking the internal `_save_grub_configuration`, `_update`, and `_update_config` functions. But those are implementation details, so if we change the name or implementation of any of those methods, this test will break. And that's fragile. That's why I advocate mocking as little as possible, and -- as much as possible -- only testing the external contract, not the implementation details. On the other hand, we probably do need to mock `subprocess.check_call` because it's an external service/program.

In any case, those are my thoughts. I'm not going to block this PR because the testing is not too my taste. We'll see if the tests end up being fragile and we can always change the testing approach later. And we can keep discussing what "unit testing" means over a beer in Riga. :-)

benhoyt (approved):

benhoyt:
@rgildein Approved now. I'm not quite sure what you mean by "Can me moved CI to runner" -- can you clarify and/or propose that in the GitHub workflow YAML?

rgildein:
> @rgildein Approved now. I'm not quite sure what you mean by "Can me moved CI to runner" -- can you clarify and/or propose that in the GitHub workflow YAML?

If you check this [line](https://github.com/canonical/operator-libs-linux/actions/runs/5798482345/job/15716332464?pr=104#step:5:8) from logs form Integrations tests, it is saying that on that runner we could not create VM machine. I though that we are using self hosted runner on which is not possible to do this, but apparently this is an issue of GitHub runner where nested virtualization is disabled. I'll check how we can enable it.

benhoyt:
@rgildein Per your Mattermost message, you say:

> The issue is that GitHub runners do not allow nested virtualization and the gtub lib require to be run on VM and not container.
Options:
>
> - We can run only Ubuntu integration tests directly on the runner VM with run: tox -e integration-ubuntu. There is no option to use CentOS. e.g. using runs-on: centos-latest
> - We can run CI on self-hosted machine, which can use lxc launch --vm ....
> - We can run CI on larger GitHub runner [1](https://github.blog/changelog/2023-02-23-hardware-accelerated-android-virtualization-on-actions-windows-and-linux-larger-hosted-runners/), where it's possible to enable KVM. However it's billed.
> - We do skip for those functional tests if they are run on container machine for now and test it only manually.

I think we should do option 1 for now -- run Ubuntu integration tests directly on a VM runner. Fairly simple, and gets us at least some automated testing for this.

rgildein:
The linting error was fixed in #105.

tests/integration/test_grub.py
```diff
@@ -0,0 +1,201 @@
...
+
+import pytest
+from charms.operator_libs_linux.v0 import grub
+
+logger = logging.getLogger(__name__)
+
+
+@pytest.fixture(autouse=True)
+def clean_configs():
+    """Clean main and charms configs after each test."""
```
benhoyt, line 15: I'm not totally up on pytest fixtures, but won't this run *before* each test? Do we instead want a [`yield` fixture](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended) so that we can run code after the test?

```python
@pytest.fixture(autouse=True)
def clean_configs():
    yield  # run test
    # grub cleanup here
```

tests/integration/test_grub.py
```diff
@@ -0,0 +1,201 @@
...
+    grub_conf_1 = grub.Config("test-charm-1")
+    grub_conf_1.update(config_1)
+    assert config_1 == grub._load_config(grub_conf_1.path)
+
+    # configure charm 2
+    grub_conf_2 = grub.Config("test-charm-2")
+    with pytest.raises(grub.ValidationError):
+        grub_conf_2.update(config_2)
+
+    assert grub_conf_2.path.exists() is False
```
benhoyt, line 149: `assert not grub_conf_2.path.exists()` would be the Pythonic way to spell this. (Similar elsewhere.)

benhoyt (approved):
I haven't gone through all the tests with a fine-toothed comb, but this looks reasonable. A couple of very minor comments, but approving.

benhoyt (approved):

benhoyt:
The code looks good. Looks like the Ubuntu integration tests are failing / timing out, though.

rgildein:
> The code looks good. Looks like the Ubuntu integration tests are failing / timing out, though.

The tests are not failing, but the "Setup operator environment" is failing. I think it's due some change in system config. We can check all tests and properly clean any configuration or simply run `juju-systemd-notices` first or even better move that to separate job. What do you think? To me the best option is to move it to separate job.

benhoyt:
For the record, @NucciTheBoss said:

> I wouldn't be surprised if the grub functional tests modified something that prevented lxd from restarting the containers correctly. For example, if you don't clean up your NFS kernel server running inside an LXD container, lxc/lxd commands will hang for most of eternity unless. That seems to have been what happened with this job https://github.com/canonical/operator-libs-linux/actions/runs/6025011553/job/16344855206

So we should probably get to the bottom of that at some point. In the meantime, I'm going to merge this. We can always improve / isolate CI later.

PR #105, opened 2023-08-15
fix comparing types and use isinstance

rgildein:
Replace types comparison with usage of `isinstance` function.

### DIFF
--- a/tests/unit/test_passwd.py
+++ b/tests/unit/test_passwd.py
@@ -21,7 +21,7 @@ class TestPasswd(TestCase):
 
     def test_user_exists_invalid_input(self):
         with self.assertRaises(TypeError):
-            print(passwd.user_exists(True))
+            passwd.user_exists(True)
 
     @patch("charms.operator_libs_linux.v0.passwd.pwd.getpwnam")
     def test_user_exists_false(self, getpwnam):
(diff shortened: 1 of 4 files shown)


tests/unit/test_passwd.py
```diff
@@ -21,7 +21,7 @@ def test_user_exists_by_uid_true(self, getpwuid):
 
     def test_user_exists_invalid_input(self):
         with self.assertRaises(TypeError):
-            passwd.user_exists(True)
+            print(passwd.user_exists(True))
```
benhoyt, line 24: Is this left-over debugging?

benhoyt (changes requested):
Won't the patch versions need incrementing due to these changes?

benhoyt:
Other than that, though, this looks good to me -- thanks!

NucciTheBoss:
This pull request should fix the failing lint tests in https://github.com/canonical/operator-libs-linux/pull/106

tests/unit/test_passwd.py
benhoyt, line 24:
```suggestion
            passwd.user_exists(True)
```

benhoyt (approved):

tests/unit/test_passwd.py
rgildein: Yes, sorry for that

PR #106, opened 2023-08-17
Enable systemd service methods to accept multiple args

NucciTheBoss:
Hello - this pull request extends the service methods in the systemd charm library to now accept multiple arguments. I made this change so that when charm authors need to pass extra parameters to `systemctl`, they do not need to directly call the private `_systemctl` method and can instead pass the extra arguments to via the public `service_*` methods.

For example, say you have a service that takes several seconds to start. Using `service_start(...)` in its currently implementation will cause Python to block until systemctl returns an exit code (`Popen.wait()`). This is inconvenient especially if you have a service that could take 30+ seconds to start. With the implementation this pull request proposes, you can pass the `--no-block` option to `service_start(...)` and let the charm continue on its way while the systemd starts the service in the background.

### Notes

* I haven't touched the _LIBAPI_ or _LIBPATCH_ versions. I will update them after reviews/approval.
* Public method signatures are the same so we shouldn't need to update the _LIBAPI_ version of the systemd charmlib. I did have to update the unit tests however because I changed the back-end implementation of `_systemctl(...)`. It is now similar to how `_dnf(...)` is implemented in the dnf charmlib.

### DIFF
--- a/lib/charms/operator_libs_linux/v1/systemd.py
+++ b/lib/charms/operator_libs_linux/v1/systemd.py
@@ -53,7 +53,6 @@ __all__ = [  # Don't export `_systemctl`. (It's not the intended way of using th
 
 import logging
 import subprocess
-from typing import Optional
 
 logger = logging.getLogger(__name__)
 
--- a/lib/charms/operator_libs_linux/v1/systemd.py
+++ b/lib/charms/operator_libs_linux/v1/systemd.py
@@ -36,10 +36,15 @@ success = service_reload("nginx", restart_on_failure=True)
 
 """
 
+import functools
 import logging
 import subprocess
 
 __all__ = [  # Don't export `_systemctl`. (It's not the intended way of using this lib.)
+    "daemon_reload",
+    "service_disable",
+    "service_enable",
+    "service_failed",
     "service_pause",
     "service_reload",
     "service_restart",
@@ -47,7 +52,6 @@ __all__ = [  # Don't export `_systemctl`. (It's not the intended way of using th
     "service_running",
     "service_start",
     "service_stop",
-    "daemon_reload",
 ]
 
 logger = logging.getLogger(__name__)
@@ -66,112 +70,50 @@ LIBPATCH = 3
 class SystemdError(Exception):
     """Custom exception for SystemD related errors."""
 
-    pass
 
-
-def _popen_kwargs():
-    return {
-        "stdout": subprocess.PIPE,
-        "stderr": subprocess.STDOUT,
-        "bufsize": 1,
-        "universal_newlines": True,
-        "encoding": "utf-8",
-    }
-
-
-def _systemctl(
-    sub_cmd: str, service_name: str = None, now: bool = None, quiet: bool = None
-) -> bool:
-    """Control a system service.
+def _systemctl(*args) -> bool:
+    """Control a system service using systemctl.
 
     Args:
-        sub_cmd: the systemctl subcommand to issue
-        service_name: the name of the service to perform the action on
-        now: passes the --now flag to the shell invocation.
-        quiet: passes the --quiet flag to the shell invocation.
+        *args: Arguments to pass to systemctl.
     """
-    cmd = ["systemctl", sub_cmd]
-
-    if service_name is not None:
-        cmd.append(service_name)
-    if now is not None:
-        cmd.append("--now")
-    if quiet is not None:
-        cmd.append("--quiet")
-    if sub_cmd != "is-active":
-        logger.debug("Attempting to {} '{}' with command {}.".format(cmd, service_name, cmd))
-    else:
-        logger.debug("Checking if '{}' is active".format(service_name))
-
-    proc = subprocess.Popen(cmd, **_popen_kwargs())
-    last_line = ""
-    for line in iter(proc.stdout.readline, ""):
-        last_line = line
-        logger.debug(line)
-
-    proc.wait()
-
-    if proc.returncode < 1:
-        return True
-
-    # If we are just checking whether a service is running, return True/False, rather
-    # than raising an error.
-    if sub_cmd == "is-active" and proc.returncode == 3:  # Code returned when service not active.
-        return False
-
-    if sub_cmd == "is-failed":
-        return False
-
-    raise SystemdError(
-        "Could not {}{}: systemd output: {}".format(
-            sub_cmd, " {}".format(service_name) if service_name else "", last_line
-        )
+    cmd = ["systemctl", *args]
+    logger.debug(f"Executing command: {cmd}")
+    proc = subprocess.run(
+        [*cmd],
+        stdout=subprocess.PIPE,
+        stderr=subprocess.STDOUT,
+        text=True,
+        bufsize=1,
+        encoding="utf-8"
     )
+    logger.debug(proc.stdout)
+    if proc.returncode == 1 and "is-failed" in cmd:
+        return False
+    elif proc.returncode == 3 and "is-active" in cmd:
+        return False
+    elif proc.returncode == 0:
+        return True
+    else:
+        raise SystemdError(f"Command {cmd} failed. systemctl output:\n{proc.stdout}")
 
 
-def service_running(service_name: str) -> bool:
-    """Determine whether a system service is running.
-
-    Args:
-        service_name: the name of the service to check
-    """
-    return _systemctl("is-active", service_name, quiet=True)
-
-
-def service_failed(service_name: str) -> bool:
-    """Determine whether a system service has failed.
-
-    Args:
-        service_name: the name of the service to check
-    """
-    return _systemctl("is-failed", service_name, quiet=True)
-
-
-def service_start(service_name: str) -> bool:
-    """Start a system service.
-
-    Args:
-        service_name: the name of the service to start
-    """
-    return _systemctl("start", service_name)
-
-
-def service_stop(service_name: str) -> bool:
-    """Stop a system service.
-
-    Args:
-        service_name: the name of the service to stop
-    """
-    return _systemctl("stop", service_name)
-
-
-def service_restart(service_name: str) -> bool:
-    """Restart a system service.
-
-    Args:
-        service_name: the name of the service to restart
-    """
-    return _systemctl("restart", service_name)
+service_running = functools.partial(_systemctl, "--quiet", "is-active")
+service_running.__doc__ = "Determine whether a system service is running."
+service_failed = functools.partial(_systemctl, "--quiet", "is-failed")
+service_failed.__doc__ = "Determine whether a system service has failed."
+service_start = functools.partial(_systemctl, "start")
+service_start.__doc__ = "Start a system service."
+service_stop = functools.partial(_systemctl, "stop")
+service_stop.__doc__ = "Stop a system service."
+service_restart = functools.partial(_systemctl, "restart")
+service_restart.__doc__ = "Restart a system service."
+service_enable = functools.partial(_systemctl, "enable")
+service_enable.__doc__ = "Enable a system service."
+service_disable = functools.partial(_systemctl, "disable")
+service_disable.__doc__ = "Disable a system service."
+daemon_reload = functools.partial(_systemctl, "daemon-reload")
+daemon_reload.__doc__ = "Reload systemd manager configuration."
 
 
 def service_reload(service_name: str, restart_on_failure: bool = False) -> bool:
@@ -179,14 +121,14 @@ def service_reload(service_name: str, restart_on_failure: bool = False) -> bool:
 
     Args:
         service_name: the name of the service to reload
-        restart_on_failure: boolean indicating whether to fallback to a restart if the
+        restart_on_failure: boolean indicating whether to fall back to a restart if the
           reload fails.
     """
     try:
         return _systemctl("reload", service_name)
     except SystemdError:
         if restart_on_failure:
-            return _systemctl("restart", service_name)
+            return service_restart(service_name)
         else:
             raise
 
@@ -199,13 +141,13 @@ def service_pause(service_name: str) -> bool:
     Args:
         service_name: the name of the service to pause
     """
-    _systemctl("disable", service_name, now=True)
+    service_disable("--now", service_name)
     _systemctl("mask", service_name)
 
     if not service_running(service_name):
         return True
 
-    raise SystemdError("Attempted to pause '{}', but it is still running.".format(service_name))
+    raise SystemdError(f"Attempted to pause '{service_name}', but it is still running.")
 
 
 def service_resume(service_name: str) -> bool:
@@ -217,14 +159,9 @@ def service_resume(service_name: str) -> bool:
         service_name: the name of the service to resume
     """
     _systemctl("unmask", service_name)
-    _systemctl("enable", service_name, now=True)
+    service_enable("--now", service_name)
 
     if service_running(service_name):
         return True
 
-    raise SystemdError("Attempted to resume '{}', but it is not running.".format(service_name))
-
-
-def daemon_reload() -> bool:
-    """Reload systemd manager configuration."""
-    return _systemctl("daemon-reload")
+    raise SystemdError(f"Attempted to resume '{service_name}', but it is not running.")
--- a/lib/charms/operator_libs_linux/v1/systemd.py
+++ b/lib/charms/operator_libs_linux/v1/systemd.py
@@ -23,6 +23,7 @@ systemd commands, with a few shortcuts baked in. For example, service_pause and
 service_resume with run the mask/unmask and enable/disable invocations.
 
 Example usage:
+
 ```python
 from charms.operator_libs_linux.v0.systemd import service_running, service_reload
 
@@ -33,12 +34,7 @@ if not service_running("mysql"):
 # Attempt to reload a service, restarting if necessary
 success = service_reload("nginx", restart_on_failure=True)
 ```
-
 """
-
-import functools
-import logging
-import subprocess
 
 __all__ = [  # Don't export `_systemctl`. (It's not the intended way of using this lib.)
     "daemon_reload",
@@ -53,6 +49,10 @@ __all__ = [  # Don't export `_systemctl`. (It's not the intended way of using th
     "service_start",
     "service_stop",
 ]
+
+import logging
+import subprocess
+from typing import Optional
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,127 +70,65 @@
...
-    """
-    return _systemctl("restart", service_name)
+    elif proc.returncode == 0:
+        return True
+    else:
+        raise SystemdError(f"Command {cmd} failed. systemctl output:\n{proc.stdout}")
+
+
+service_running = functools.partial(_systemctl, "--quiet", "is-active")
+service_running.__doc__ = "Determine whether a system service is running."
```
benhoyt, line 102: So I'm okay with using functools.partial, and thanks for setting the docstring. However, I'm concerned that we're losing the type annotation here. It looks like `functools.partial` does copy across the type signature (see `inspect.signature`), but in this case `_systemctl` doesn't use type annotations, so the partial functions won't either.

Relatedly, what will the docs look like here? https://charmhub.io/operator-libs-linux/libraries/systemd#service_running ... I think we need to keep these informative (and with type annotations). I notice that even as it stands they don't have return type annotations (`bool` for some functions, `None` for others?), so maybe we can address that while we're at it.

This *may* mean writing out all the functions rather than using `functools.partial`, but I think if so, that's an okay trade-off. Better to make it nice for our users every time (docs and type annotations) rather than nice for us once.

Some of these functions also return `bool` (I don't see why funcs like `service_start` should return bool as I believe they raise an exception on failure, but we're probably locked into the existing type signatures unless we release a major version bump).

One more thing: I realize you're just copying the existing docstrings, but it's currently unclear what these first two functions do after "determining whether" the thing occurred (they return the result as a bool). I like the Go style of saying "Report whether ...", but "Return True if ..., otherwise False" is okay too.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,127 +70,65 @@
...
+service_enable = functools.partial(_systemctl, "enable")
+service_enable.__doc__ = "Enable a system service."
+service_disable = functools.partial(_systemctl, "disable")
+service_disable.__doc__ = "Disable a system service."
+daemon_reload = functools.partial(_systemctl, "daemon-reload")
+daemon_reload.__doc__ = "Reload systemd manager configuration."
 
 
 def service_reload(service_name: str, restart_on_failure: bool = False) -> bool:
     """Reload a system service, optionally falling back to restart if reload fails.
```
benhoyt, line 120: Related to the above, what does the return value of this function mean? Or will it always return True or raise an error?

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -199,13 +141,13 @@ def service_pause(service_name: str) -> bool:
...
     """
-    _systemctl("disable", service_name, now=True)
+    service_disable("--now", service_name)
     _systemctl("mask", service_name)
 
     if not service_running(service_name):
         return True
 
-    raise SystemdError("Attempted to pause '{}', but it is still running.".format(service_name))
+    raise SystemdError(f"Attempted to pause '{service_name}', but it is still running.")
```
benhoyt, line 150: Probably won't matter here as I doubt service names have quotes in them, but `{service_name!r}` is better than `'{service_name}'` as it uses the repr (which includes quotes and escaping).

tests/unit/test_systemd.py
```diff
@@ -9,38 +10,41 @@
...
+            mock_run.side_effect = tuple(side_effects)
 
-            return mock_popen, systemd._popen_kwargs()
+            return mock_run, {
+                "stdout": subprocess.PIPE,
+                "stderr": subprocess.PIPE,
+                "text": True,
+                "bufsize": 1,
+                "encoding": "utf-8",
+            }
```
benhoyt, line 47: Appreciate this, thanks! Good to avoid `_private` funcs in tests, and to spell things out separately, otherwise you're just reusing (potentially) buggy data in the test.

benhoyt (changes requested):
Thanks for this. Direction looks good, but a couple of suggestions and questions. I've also pinged rgildein about https://github.com/canonical/operator-libs-linux/pull/105

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,127 +70,65 @@
...
-    """Restart a system service.
-
-    Args:
-        service_name: the name of the service to restart
-    """
-    return _systemctl("restart", service_name)
+    elif proc.returncode == 0:
+        return True
+    else:
+        raise SystemdError(f"Command {cmd} failed. systemctl output:\n{proc.stdout}")
```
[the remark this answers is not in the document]
NucciTheBoss: Hey Ben! So I mostly had the evaluation of the the return codes like this in `_systemctl(...)` because that's how it was in the earlier version of the `_systemctl(...)` :sweat_smile:  

I agree that it feels weird that we are following an if-else chain within `_systemctl` and modifying the return value based on specific use-cases. I like the idea of using a `check=False` kwarg to direct `_systemctl(...)` to raise an error or not. I do believe that _systemctl should just return the return code since it should handle logging and raising the SystemdError itself.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Either way using `{service_name!r}` is good insurance that if someone decides to make the service names have quotes, it won't bork the charmlib and require us to issue another update. Let me make this change.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Which function are we talking about? My view on GitHub shows multiple functions unfortunately... If we are discussing `service_reload(...)` the return value is True and an error will be raised if `systemctl reload ...` and/or `systemctl restart ...` fail (returncode != 0). True just means that the operation to reload and/or restart a service using systemctl was successful. 

This bool value returned by the `service_reload(...)` function is a holdover from the original implementation of `_systemctl(...)` where False was returned if the "is-active" or "is-failed" did not return the expected exit code to indicate that the targeted service is not running or has failed.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: I can change the docstrings for `service_running(...)` and `service_failed(...)` to be more descriptive than they are now. Do not want to create any sense of ambiguity!

I believe we can type `*args` in `_systemctl(...)` by using the [Unpack](https://docs.python.org/3/library/typing.html?highlight=unpack#typing.Unpack) type provided by the `typing_extensions` module. Unfortunately, our suspicion that Charmhub will not render the partial functions in the charmlib documentation is like true, so I will need to switch them back to the tradition `def service_start(...) -> bool:` format.

Yes, I agree that it doesn't make sense for methods like `service_start(...)` and `service_stop(...)` to return a bool value, but this is also a holdover from the original implementation of the `_systemctl(...)` method where it was being used to evaluate the output of "is-active" and "is-failed". We could change it, but we will need to release a new major version rather than a patch version since I do not know if there are other charmers who depend upon this behavior.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -36,18 +36,22 @@
 
 """
 
+import functools
 import logging
 import subprocess
 
 __all__ = [  # Don't export `_systemctl`. (It's not the intended way of using this lib.)
```
NucciTheBoss, line 43: While I am in the neighborhood, I noticed the having `__all__` declared after the stdlib module imports violates [PEP8](https://peps.python.org/pep-0008/#module-level-dunder-names). `__all__` should come before the module imports but after the module docstring. Easy fix!

NucciTheBoss: Fixed in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/af8c1fb3b34fbfc19b74b20d4210dd45c7eb07db

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/af8c1fb3b34fbfc19b74b20d4210dd45c7eb07db

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Ripped off the band-aid :adhesive_bandage: and changed annotations for methods that should not return anything to `None` in https://github.com/canonical/operator-libs-linux/pull/106/commits/af8c1fb3b34fbfc19b74b20d4210dd45c7eb07db.

Now, if an error is encountered, a `SystemdError` will be raised. The returned exit code of `0` by _systemctl is ignored in the public methods, but _systemctl itself will log the exit code, command, and stdout when in debug mode.

lib/charms/operator_libs_linux/v1/systemd.py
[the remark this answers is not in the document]
NucciTheBoss: `service_running(...)` and `service_failed(...)` now evaluate the return codes themselves to determine whether `True`/`False` should be returned rather than relying on _systemctl to filter the `subprocess.CompletedProcess` returned by `subprocess.run`. This change makes the implementation of `_systemctl(...)` much simpler :sweat_smile: 

Change made in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/af8c1fb3b34fbfc19b74b20d4210dd45c7eb07db

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: So I switched back to using proper function definitions rather than using `functools.partial` because functions defined using `functools.partial` will not be rendered properly in Charmhub's charmlib documentation. For `*args`, I just typed it as `str` because I did not want to bring in `typing_extensions` as a dependency. I also updated the docstrings of `service_running(...)` and `service_failed(...)` to be more descriptive.

I also changed the signatures for methods like `service_start(...)`, `service_stop(...)`, etc to return `None` rather than a bool. It makes sense to not return anything since `service_start(...)` should throw an error if a service fails to start. The charm author should consciously chose to ignore that their service has failed to start correctly by handling the raised `SystemdError`. Only thing is that we should determine if this change warrants bumping the LIBAPI version of the systemd library. Let me know what you think.

NucciTheBoss:
BAH - integration tests need to be updated since `service_start(...)`, `service_stop(...)`, etc no longer return boolean values.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,149 +71,146 @@
...
+    Raises:
+        SystemdError:
+            Raised if returncode != 0 and check is True.
+            None will be returned instead of a returncode.
+    """
+    cmd = ["systemctl", *args]
+    logger.debug(f"Executing command: {cmd}")
+    try:
+        proc = subprocess.run(
+            [*cmd],
```
benhoyt, line 92: I think `[*cmd]` is equivalent to just `cmd`.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,149 +71,146 @@
...
-def _popen_kwargs():
-    return {
-        "stdout": subprocess.PIPE,
-        "stderr": subprocess.STDOUT,
-        "bufsize": 1,
-        "universal_newlines": True,
-        "encoding": "utf-8",
-    }
+    Args:
+        *args: Arguments to pass to systemctl.
```
benhoyt, line 79: For this to render properly in the docs, is the proper annotation for this `args:` or `*args:`? I'm guessing the former, but I'm not sure.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,149 +71,146 @@
...
+        check:
+            Check the output of the systemctl command. Default: False.
+
+    Raises:
+        SystemdError:
+            Raised if returncode != 0 and check is True.
+            None will be returned instead of a returncode.
```
benhoyt, line 86: When will `None` be returned from this function? From looking at the code structure below, it looks like it can only return an `int` exit code *or* raise `SystemdError`. If so, we should tweak this docstring and change this function to return type `int` instead of `Optional[int]`.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,149 +71,146 @@
...
-    proc.wait()
+def service_failed(service_name: str) -> bool:
+    """Report whether a system service has failed.
 
-    if proc.returncode < 1:
+    Args:
+        service_name: the name of the service to check
+    """
+    # If returncode is 0, this means that the service has failed.
+    if _systemctl("--quiet", "is-failed", service_name) == 0:
```
benhoyt, line 130: Weird but true that ==0 means failed in this case!

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,149 +71,146 @@
...
-    if service_name is not None:
-        cmd.append(service_name)
-    if now is not None:
-        cmd.append("--now")
-    if quiet is not None:
-        cmd.append("--quiet")
-    if sub_cmd != "is-active":
-        logger.debug("Attempting to {} '{}' with command {}.".format(cmd, service_name, cmd))
+    # If returncode is 0, this means that is service is active.
+    if _systemctl("--quiet", "is-active", service_name) == 0:
```
benhoyt, line 117: Simpler (and arguably clearer?) to just do `return _systemctl("--quiet", "is-active", service_name) == 0`. Similar below.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,149 +71,146 @@
...
-    raise SystemdError(
-        "Could not {}{}: systemd output: {}".format(
-            sub_cmd, " {}".format(service_name) if service_name else "", last_line
-        )
-    )
-
 
-def service_running(service_name: str) -> bool:
-    """Determine whether a system service is running.
+def service_start(*args: str) -> None:
```
benhoyt, line 136: Yeah, on reflection I think we probably shouldn't make the change to the return type without a major version bump (and I don't think it's worth doing that). So what about doing something like this for all of these functions.

```python
def service_start(*args: str) -> bool:
    """Start a system service.

    Args:
        args: Name (or names) of services to start.

    Returns:
        On success, this function always returns True (for historical reasons).

    Raises:
        SystemdError: on error calling systemctl.
"""
    _systemctl("start", *args, check=True)
    return True
```

benhoyt (changes requested):
Thanks for the restructure and docs improvement. This is very close now, but per my comment I think we should avoid changing the return types, just in case: `None` is the opposite -- in the "truthy" sense -- of the current value of True that's being returned, so this could well break people.

lib/charms/operator_libs_linux/v1/systemd.py
benhoyt: Good changes, thanks. However, I do think we should switch back to these returning true to avoid a major version bump. See my comments below.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Hmm... from a quick search online, there does not seem to be clear consensus on the proper annotation for `*args` in docstrings. Anecdotally, the docstrings for the `dnf` charm library seem to render fine on Charmhub: https://charmhub.io/operator-libs-linux/libraries/dnf. Several of the methods in the dnf charm library use `*packages` in the docstring for documenting the `*packages` argument.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Sounds good - I can change it back to returning `True`. Since _systemctl(...) can either return an exit code of 0 or raise a `SystemdError`, we can just have the return line be `return _systemctl("start", *args, check=True) == 0`. I don't like the idea of returning `True` just using `return True` to ensure that we don't break compatibility because it appears that we are just ignoring what `_systemctl(...)` does to the underlying machine.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: In my haste I said that `None` is returned if check=True and the returncode!=0, however, this is not true on second glance. I will go ahead and tweak the docstring.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Switched back to returning bools in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/c951f68b8fe18cbab6fc69858e838283063c1e48

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Implemented in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/c951f68b8fe18cbab6fc69858e838283063c1e48

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Revised in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/c951f68b8fe18cbab6fc69858e838283063c1e48

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/106/commits/c951f68b8fe18cbab6fc69858e838283063c1e48

lib/charms/operator_libs_linux/v1/systemd.py
benhoyt: Interesting. One thing that's weird is that they render as:

```python
def install()
```

Whereas I would have expected something like this:

```python
def install(*packages)
```

As is, it looks like the function doesn't take any arguments, until you scan down to the argument section, and then it's confusing. Maybe that's because of the `*packages` name? Or maybe that's some other issue. I'm not sure what tool we're using to generate that on Charmhub.io.

lib/charms/operator_libs_linux/v1/systemd.py
```diff
@@ -66,165 +70,219 @@
...
+    Returns:
+        True if service is marked as failed; False if not.
     """
-    return _systemctl("is-failed", service_name, quiet=True)
+    # If returncode is 0, this means that the service has failed.
+    return _systemctl("--quiet", "is-failed", service_name) == 0
 
 
-def service_start(service_name: str) -> bool:
+def service_start(*args: str) -> bool:
```
benhoyt, line 135: This should probably be named `*service_names` rather than the generic `*args` -- when you're just using `*args` or `**kwargs` as a pass-through and you don't know the meaning, use the generic terms, but when the argument has specific meaning (the services names in this case) it's best to use a proper name.

benhoyt (approved):
Looks good to me! Just one nit comment, but approving.

lib/charms/operator_libs_linux/v1/systemd.py
NucciTheBoss: Hmm... normally I would agree with you that `*<parameter_name>` should capture the meaning of the positional arguments being passed to the method, however, I don't find that `*service_names` is a good fit here. Reason saying is that `service_start` | `service_stop` | `service_restart` can (and should) accept more than just service names as position arguments.

For example, the I have a use case where I need to pass the `--no-block` flag to `systemctl start ...` because the service I am working with is `Type=forking` and the parent process can take potentially take ~15 minutes to return control to systemd. Long story... but I feel that if we change `*args` -> `*service_names`, charm authors consuming this lib may not realize that they are allowed to pass flags to `systemctl` as well as service names. Therefore, I think that `*args` is a better parameter name in this situation since the `service_*` methods now allow you to pass generic arguments to `systemctl` via `subprocess.run([...])`.

benhoyt: Ah, I missed that nuance -- makes sense!

PR #108, opened 2023-08-23
feat: Add juju systemd notices charm library

NucciTheBoss:
Hello hello -

This pull request proposes adding the machine charm library __juju_systemd_notices__ to operator-libs-linux. The juju_systemd_notices charm library is already used within the HPC team, but we are looking for an upstream home (+ reviews) for this charm library so that other charm teams can make use of it.

This pull request itself adds:

* The juju_systemd_notices charm library itself.
* Unit tests for the juju_systemd_notices charm library.
* Integration tests for the juju_systemd_notices charm library. 

You can view the juju_systemd_notices charm library in action within the slurmd-operator here: https://github.com/omnivector-solutions/slurmd-operator.

#### Miscellaneous

I only modified files particular to juju_systemd_notices, but a few extra commits are coming up because I needed to rebase my feature branch where I was developing the library. I had a couple `chore(wip)` commits so I could share my current progress with others, but I squashed them into one commit before opening this pull request.


NucciTheBoss:
Hmm... seems like there is a discrepancy between `with (... patch(...))` in Python 3.8 versus 3.10. Shouldn't be too hard to fix.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+
+
+class ServiceStoppedEvent(EventBase):
+    """Event emitted when service has stopped."""
+
+
+class SystemdNotices:
+    """Observe systemd services on your machine base."""
+
+    def __init__(self, charm: CharmBase, services: Union[str, List[str]]) -> None:
```
benhoyt, line 166: I usually prefer simpler signatures instead of "either str or list", so either `*services: str` (probably best here) or `services: List[str]`. "There should be one ... obvious way to do it." PEP 20 :-)

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+        for service in self._services:
+            self._charm.on.define_event(f"service_{service}_started", ServiceStartedEvent)
+            self._charm.on.define_event(f"service_{service}_stopped", ServiceStoppedEvent)
+
+    def subscribe(self) -> None:
+        """Subscribe charmed operator to observe status of systemd services."""
+        _logger.debug(f"Generating systemd notice hooks for {self._services}")
+        for service in self._services:
+            Path(f"hooks/service-{service}-started").symlink_to(f"{Path.cwd()}/dispatch")
+            Path(f"hooks/service-{service}-stopped").symlink_to(f"{Path.cwd()}/dispatch")
```
benhoyt, line 183: It seems weird to use the current working directory here (global state). Is there a way to calculate this from the charm directory? `Framework` has a `charm_dir` attribute, for example. Similar elsewhere.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+        self._services = [services] if isinstance(services, str) else services
+        unit_name = self._charm.unit.name.replace("/", "-")
+        self._service_file = Path(f"/etc/systemd/system/juju-{unit_name}-systemd-notices.service")
+
+        _logger.debug(f"Attaching systemd notice events to charm {self._charm.__class__.__name__}")
+        for service in self._services:
+            self._charm.on.define_event(f"service_{service}_started", ServiceStartedEvent)
+            self._charm.on.define_event(f"service_{service}_stopped", ServiceStoppedEvent)
+
+    def subscribe(self) -> None:
```
benhoyt, line 178: Not sure, but should this be called `start` to parallel `stop`?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+LIBID = "HPCTEAMUSEONLY"
+LIBAPI = 0
+LIBPATCH = 1
+PYDEPS = ["dbus-fast>=1.90.2"]
+
+_logger = logging.getLogger(__name__)
+_juju_unit = None
+_service_states = {}
+_service_hook_regex_filter = re.compile(r"service-(?P<service>[\w\\:-]*)-(?:started|stopped)")
+_DBUS_CHAR_MAPPINGS = MappingProxyType(
```
benhoyt, line 117: It's pretty unusual to use this. I realize it creates a read-only view of a dict, but this is a private constant -- let's just KISS and use a regular dict.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+    {
+        "_5f": "_",
```
benhoyt, line 119: Maybe just add an inline comment here: `# _ must be first`

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+    Args:
+        service: The name of the service which has changed state.
+        state: The state of the service.
+    """
+    if service.endswith(".service"):
+        service = service[0 : -len(".service")]
+    if state == "active":
+        event_name = "started"
+    else:
+        event_name = "stopped"
```
benhoyt, line 310: Inline if/else is nice for this: `event_name = 'started' if state == 'active' else 'stopped'`

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+    Any other hooks are ignored and not loaded into the set of services
+    that should be watched. Upon finding a service hook it's current ActiveState
+    will be queried from systemd to determine it's initial state.
+    """
+    global _juju_unit
+    hooks_dir = Path.cwd() / "hooks"
+    _logger.info(f"Loading services from hooks in {hooks_dir}")
+
+    if not hooks_dir.exists():
+        _logger.warning(f"Hooks dir {hooks_dir} does not exist.")
```
benhoyt, line 370: Should this be an error/exception? Isn't it essentially fatal and prevents the operation of this module?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+        return
+
+    watched_services = []
+    # Get service-{service}-(started|stopped) hooks defined by the charm.
+    for hook in filter(lambda p: _service_hook_regex_filter.match(p.name), hooks_dir.iterdir()):
```
benhoyt, line 375: We're doing the `match` call twice. Besides, I think this would be simpler and clearer as:

```python
for hook in hooks_dir.iterdir():
    match = _service_hook_regex_filter.match(hook.name)
    if match:
        watched_services.append(match.group("service"))
```

Or -- I'm not the biggest fan of the walrus operator (Python 3.8+), but it does work pretty well here:

```python
watched = [match.group("service") for h in hooks_dir.iterdir()
           if (match := _service_hook_regex_filter.match(h.name))]
```

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+    """
+    _logger.debug(
+        f"Received message: path: {msg.path}, interface: {msg.interface}, member: {msg.member}"
+    )
+    service = _dbus_path_to_name(msg.path)
+    properties = msg.body[1]
+    if "ActiveState" not in properties:
+        return False
+
+    global _service_states
```
benhoyt, line 278: Are there any concurrency issues with global state being used/updated in this way? Or is it sequential access due to how async works? I'm not super familiar with it, but I presume it runs everything on a single-threaded loop.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+    if state == "active":
+        event_name = "started"
+    else:
+        event_name = "stopped"
+    hook = f"service-{service}-{event_name}"
+    cmd = ["/usr/bin/juju-exec", _juju_unit, f"hooks/{hook}"]
+
+    _logger.debug(f"Invoking hook {hook} with command: {' '.join(cmd)}")
+    process = await asyncio.create_subprocess_exec(
+        *cmd,
```
benhoyt, line 316: Total nit, but this seems like it would be clearer on one line.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+        console_handler.setLevel(logging.DEBUG)
+    else:
+        _logger.setLevel(logging.INFO)
+        console_handler.setLevel(logging.DEBUG)
+    _logger.addHandler(console_handler)
+
+    # Intentionally set as global.
+    global _juju_unit
+    _juju_unit = args.unit
+    if not _juju_unit:
```
benhoyt, line 476: I think this check is automatically done by `ArgumentParser` for positional arguments like "unit".

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+    """
+    stop_event = asyncio.Event()
+    loop = asyncio.get_event_loop()
+    loop.add_signal_handler(signal.SIGINT, stop_event.set)
+    loop.add_signal_handler(signal.SIGTERM, stop_event.set)
+    # The event loop must be early bound to the lambda, otherwise the event loop
+    # will not exist within the lambda when the SIGHUP signal is received
+    # by the running notices daemon.
+    loop.add_signal_handler(  # pragma: no branch
+        signal.SIGHUP, lambda loop=loop: _load_services(loop)
```
benhoyt, line 427: So just to clarify how this works: systemd sends `SIGHUP` to this process to say "got something new", and then `_load_services` queries the current state using the dbus API?

.github/workflows/build-and-test.yaml
```diff
@@ -44,3 +44,10 @@ jobs:
         run: python -m pip install tox
       - name: Run integration tests
         run: tox -e integration
+      - name: Setup operator environment
+        uses: charmed-kubernetes/actions-operator@main
```
benhoyt, line 48: Why is charmed-kubernetes need here? Does it add much time to the integration tests run?

benhoyt (commented):
I haven't had time to look at the tests yet, but this looks reasonable overall, thanks. I've added a few comments/questions.

.github/workflows/build-and-test.yaml
jnsgruk: This is just a standard action that sets up charmcraft/juju/lxd etc., so not necesarrily deploy k8s in this instance.

My only question is whether it's needed - could this be covered with more simple functional tests using LXD like the other libraries?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+_daemon_reload = functools.partial(_systemctl, "daemon-reload")
+_daemon_reload.__doc__ = "Reload systemd manager configuration."
+_start_service = functools.partial(_systemctl, "start")
+_start_service.__doc__ = "Start systemd service unit."
+_stop_service = functools.partial(_systemctl, "stop")
+_stop_service.__doc__ = "Stop systemd service unit."
+_enable_service = functools.partial(_systemctl, "enable")
+_enable_service.__doc__ = "Enable systemd service."
+_disable_service = functools.partial(_systemctl, "disable")
+_disable_service.__doc__ = "Disable systemd service."
```
jnsgruk, line 152: We already have a systemd library - how do folks feel about just introducing that as a dep here and using that?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Sounds good - I'm just paranoid about folks doing things that I don't want them too such as modifying mutable objects like dict/list :sweat_smile: 

I can remove this however if we feel that using `MappingProxyType` is not necessary here.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Good idea. I think I wrote this factoid down in one of my commit messages, but it would be good to also include this as an inline comment too for future reference.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I don't love this formatting either. This was a change that `black` made after I ran `tox run -e fmt`. I tried to change it back, but `black` really preferred this formatting for some reason. Let me see if there is something that I can do to make black NOT do this :smiley:

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: :man_facepalming: - you're right... let me refactor this.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I named this method `subscribe(...)` rather than `start(...)` because it does more than just start the `juju-systemd-notices` daemon in the background. It:

1. Creates the hooks for the service start/stop events
2. Creates the `juju-systemd-notices` service file
3. Reloads the systemd manager configuration
4. Enables the daemon (so it will start even after machine reboot)
5. Starts the `juju-systemd-notices` daemon

I debated naming the method `start(...)`, but given that it does more than just start the daemon, I didn't want to create any confusion. E.g. why is this method creating a service/creating hooks off of the dispatch hook? Let me know what your thoughts are about this.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I haven't seen any issues since we don't create any separate threads or event loops. We did initially have a ThreadPoolExecutor to get around a RuntimeError we were encountering when trying to call the async event loop when there was none (synchronous -> asynchronous), but we fixed that issue by early binding the event loop to the lambda that invokes the synchronous `_load_services(...)` function.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I would prefer that personally since then we're not on the hook for maintaining two separate implementations for interfacing with `systemctl` in `operator-libs-linux`. My only hang up however (and the reason why I initially implemented these methods) is that I am unsure how we would pull in the systemd charm library as a dependency. Can we use PYDEPS for this?

While we could _tell_ folks in the documentation for this charm lib that they must also consume the systemd charm library, I am concerned that if we make this workflow manual then someone will definitely bungle it by forgetting to pull in the systemd charm library before packing their charm.

jnsgruk: Yeh, that is a bit of a thing, but it's not the end of the world - in reality it would fail right away in unit tests etc.

The Observability team did this for a while with their `JujuTopology` lib - I think so long as it's properly documented in the docstring, it's probably the way to go :)

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: SIGHUP is sent to daemons by systemd to reload the complete daemon configuration. When `juju-systemd-notices` receives the SIGHUP signal, this tells the daemon to reload the watched services based on what hooks are present in `hooks/` and enter their current state in `_service_states`. That way when new messages are received on the DBus message bus after the reload, the daemon reacts correctly instead of prematurely emitting an event because the daemon forgot the previous state of the service that it is watching.

The systemd manpage goes into more detail about signals that can be sent to services: https://manpages.ubuntu.com/manpages/jammy/en/man1/systemd.1.html#signals

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: The `juju-systemd-notices` daemon will still start successfully in the background even if there are no detected hooks. It will run, but the charm itself will not receive a `ServiceStartEvent`|`ServiceStopEvent` for the intended service. We could make this be a failure instead, but we would need to include a mechanism for ensuring that the error bubbles back up into the charm.

We can make the `juju-systemd-notices` daemon fail, but that failure won't do us much good if the charm does not have any way to react to that failure since `juju-systemd-notices` runs independently of the main charm process.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Hmm... it makes sense to use the `charm_dir` attribute in the public charm API part of the charm lib, but we cannot use it in the daemon part of the charm lib. Is there a particular concern with using `Path.cwd()` in the daemon bits of this charm lib? While we might have to worry about someone changing the current working directory inside their charm before setting up the `juju-systemd-notices` daemon, but the in the daemon bits we know that we will be in the correct directory since we set the WorkingDirectory in the service file for the daemon.

.github/workflows/build-and-test.yaml
NucciTheBoss: The reason I added this step is because I need a mini-Juju cloud to test that the `juju-systemd-notices` daemon properly feeds `ServiceStartEvent`|`ServiceStopEvent` events into a test charm and that the test charm properly handles the events. It does add additional overhead to the integration tests, but I need a Juju environment to ensure that the charm lib is working as expected in practice.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
benhoyt: Ah yeah, fair enough -- when I wrote this comment I wasn't thinking about the daemon part. I think that's fine: let's use `charm_dir` in the API part but `Path.cmd()` is fine in the daemon part.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
benhoyt: Yeah, makes sense. I'd be fine with `subscribe`. Could also consider `start_and_subscribe` -- a bit of a mouthful, but very clear.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
benhoyt: Fair enough, might as well keep it simple.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
benhoyt: Weird. I wonder if you had something else on the line before so it was wrapping? I just tried this on one line and Black doesn't seem to mind:

```
$ cat t.py
import asyncio


def foo(cmd):
    _logger.debug(f"Invoking hook {hook} with command: {' '.join(cmd)}")
    process = await asyncio.create_subprocess_exec(*cmd)
    await process.wait()
$ black --version
black, 23.7.0 (compiled: yes)
Python (CPython) 3.10.12
$ black t.py
All done! ✨ 🍰 ✨
1 file left unchanged.
```

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
benhoyt: I'd agree that would be a good trade-off if this was a complex chunk of code.

But here I'd much prefer the slight duplication to make it easier for users. As you note, charm libs can't have dependencies on each other as you (only on other PyPI libs via `PYDEPS`). As the [saying `go`es](https://go-proverbs.github.io/), "A little copying is better than a little dependency." It's easier for us to copy/maintain a few straight-forward lines of code (not much more than leftpad :-) than it will be for each consumer to remember to add the other charm lib.

Also, probably no need to copy the docstrings on these partials -- these are internal functions and pretty obvious. So that would save 5 lines of code. :-)

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Inline comment added in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Removed `MappingProxyType` in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Replaced `Path.cwd()` with `self._charm.framework.charm_dir` in the public charm API in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Refactored `__init__(...)` signature in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+        # Register services with charm. This adds the events to observe.
+        self._systemd_notices = SystemdNotices(self, ["slurmd"])
+        event_handler_bindings = {
+            self.on.install: self._on_install,
+            self.on.stop: self._on_stop,
+            self.on.service_slurmd_started: self._on_slurmd_started,
+            self.on.service_slurmd_stopped: self._on_slurmd_stopped,
+        }
+        for event, handler in event_handler_bindings.items():
+            self.framework.observe(event, handler)
```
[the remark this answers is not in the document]
NucciTheBoss: Revised in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/a61cf98ddfaf54f73b3c56c9182208052ca01032

NucciTheBoss:
Alright @benhoyt and @jnsgruk - just finished this round of updates per your feedback so far. Let me know what you think after the CI checks complete!

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
benhoyt: Just for the record, I don't think this is true:

> I believe this comma is here to prevent asyncio.create_subprocess_exec(...) from thinking the cmd list contains inputs for other values that asyncio.create_subprocess_exec(...) accepts

The comma is simply an optional trailing comma, and doesn't have any semantic meaning. See the `arguments:` rule in the [Python grammar](https://docs.python.org/3/reference/grammar.html), it has an optional `[',']` before the right parenthesis.

So `f(*cmd) is exactly equivalent to `f(*cmd,)`, the latter just has a trailing comma.

In addition, it wouldn't make sense for positional `*args` to "fill up" keyword args like `stdin=None` in this function's case.

You can show this by defining a function with the same args as `create_subprocess_exec` and then calling it with `*cmd`, and see that the other args aren't set:

```
>>> def f(program, *args, stdin=None, stdout=None, stderr=None, limit=None, **kwds):
...     print(program, args, stdin, stdout, stderr, limit, kwds)
...
>>> cmd = ['systemctl', 'start']
>>> f(*cmd)
systemctl ('start',) None None None None {}
>>> f(*cmd,)
systemctl ('start',) None None None None {}
```

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+            _logger.setLevel(logging.DEBUG)
+            console_handler.setLevel(logging.DEBUG)
+        else:
+            _logger.setLevel(logging.INFO)
+            console_handler.setLevel(logging.DEBUG)
+
+        _logger.addHandler(console_handler)
+        _logger.info("Starting juju systemd notices service")
+        asyncio.run(_juju_systemd_notices_daemon())
+    except argparse.ArgumentError:
```
benhoyt, line 470: Isn't this the default for `ArgumentParser`? To print a usage message and exit? In my experience you don't usually need this try/except in scripts because of the default.

benhoyt (approved):
Looks good, thanks! One very minor comment, but approving.

Are you able to add a proper `LIBID` or does @jnsgruk need to do that?

benhoyt:
@NucciTheBoss Two other things:

1) Would also be good to update the README with a link to this lib under Usage.
2) Integration tests are failing. Doesn't look related to this PR, but are you able to investigate? If so, it might be an intermittent failure as [this](https://github.com/canonical/operator-libs-linux/actions/runs/5996972270/job/16262466171) succeeded earlier today.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+    def subscribe(self) -> None:
+        """Subscribe charmed operator to observe status of systemd services."""
+        _logger.debug(f"Generating systemd notice hooks for {self._services}")
+        for hook in [
+            Path(f"hooks/service-{service}-{hook_type}")
+            for service in self._services
+            for hook_type in ("started", "stopped")
+        ]:
+            if hook.exists():
+                _logger.debug(f"Hook {hook.name} already exists. Skipping...")
```
jnsgruk, line 175:
```suggestion
                _logger.debug("Hook %s already exists. Skipping...", hook.name)
```

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+    """Observe systemd services on your machine base."""
+
+    def __init__(self, charm: CharmBase, services: List[str]) -> None:
+        """Instantiate systemd notices service."""
+        self._charm = charm
+        self._services = services
+        unit_name = self._charm.unit.name.replace("/", "-")
+        self._service_file = Path(f"/etc/systemd/system/juju-{unit_name}-systemd-notices.service")
+
+        _logger.debug(f"Attaching systemd notice events to charm {self._charm.__class__.__name__}")
```
jnsgruk, line 161:
```suggestion
        _logger.debug("Attaching systemd notice events to charm %s", self._charm.__class__.__name__)
```

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+
+
+def _systemctl(*args) -> None:
+    """Control systemd by via executed `systemctl ...` commands.
+
+    Raises:
+        subprocess.CalledProcessError: Raised if systemctl command fails.
+    """
+    cmd = ["systemctl", *args]
+    _logger.debug(f"systemd: Executing command {cmd}")
```
jnsgruk, line 128:
```suggestion
    _logger.debug("systemd: Executing command %s", cmd)
```

Please don't build log strings like this :) Applies throughout.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+
+    global _service_states
+    if service not in _service_states:
+        _logger.debug(f"Dropping event for unwatched service: {service}")
+        return False
+
+    curr_state = properties["ActiveState"].value
+    prev_state = _service_states[service]
+    # Drop transitioning and duplicate events
+    if curr_state.endswith("ing") or curr_state == prev_state:
```
jnsgruk, line 280: Ha, this is neat :)

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+    await process.wait()
+    if process.returncode:
+        _logger.error(
+            f"Hook command '{' '.join(cmd)}' failed with returncode {process.returncode}"
+        )
+    else:
+        _logger.info(f"Hook command '{' '.join(cmd)}' succeeded.")
+
+
+async def _get_state(bus: MessageBus, service: str) -> str:
```
jnsgruk, line 317:
```suggestion
async def _get_service_state(bus: MessageBus, service: str) -> str:
```

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: The big thing is more or less that I want to emit an exit code 2. That way if systemd reports `juju-systemd-notices` as failed, we will know that the daemon failed due to misuse of arguments rather than that it encountered a fatal error when running. After running a faulty test without the try-except block, it does look like the exit code is 2 when you do not specify the unit name, so I will make this change.

NucciTheBoss:
> 2\. Integration tests are failing. Doesn't look related to this PR, but are you able to investigate? If so, it might be an intermittent failure as [this](https://github.com/canonical/operator-libs-linux/actions/runs/5996972270/job/16262466171) succeeded earlier today.

@benhoyt sure I can take a look - seems like the `test_snap_refresh` integration test is busted. It looks like it is timing out trying to refresh the `lxd` snap to `latest/candidate`; it's hanging on waiting for lxd daemon to stop. This might not be right, but I believe it's hanging because the integration tests are using the lxd-agent inside the test container images to communicate with the images. It's probably hanging because it's waiting for the connection to close, but since we are using it to communicate with the container, it won't close. Easiest solution is to likely just change the snap that we are testing/refreshing.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Sounds good. I am going to author these commits locally rather than committing the suggestions so that I can sign the commits using my GPG key.

NucciTheBoss: Refactor in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/3bd603a65b70ec45dfbc4ceff82114739179f9c6

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/3bd603a65b70ec45dfbc4ceff82114739179f9c6

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/3bd603a65b70ec45dfbc4ceff82114739179f9c6

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,476 @@
...
+            self._charm.on.define_event(f"service_{service}_stopped", ServiceStoppedEvent)
+
+    def subscribe(self) -> None:
+        """Subscribe charmed operator to observe status of systemd services."""
+        _logger.debug(f"Generating systemd notice hooks for {self._services}")
+        for hook in [
+            Path(f"hooks/service-{service}-{hook_type}")
+            for service in self._services
+            for hook_type in ("started", "stopped")
+        ]:
```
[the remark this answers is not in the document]
NucciTheBoss: Refactored in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/3bd603a65b70ec45dfbc4ceff82114739179f9c6

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -0,0 +1,485 @@
...
+from typing import List, Union
+
+from dbus_fast.aio import MessageBus
+from dbus_fast.constants import BusType, MessageType
+from dbus_fast.errors import DBusError
+from dbus_fast.message import Message
+from ops.charm import CharmBase
+from ops.framework import EventBase
+
+LIBID = "HPCTEAMUSEONLY"
```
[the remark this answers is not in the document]
NucciTheBoss: Updated in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/3bd603a65b70ec45dfbc4ceff82114739179f9c6. Thanks for registering the lib @jnsgruk :rocket:

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Changed in commit https://github.com/canonical/operator-libs-linux/pull/108/commits/3bd603a65b70ec45dfbc4ceff82114739179f9c6. After running the integration tests successfully I decided that the comma was no longer necessary.

NucciTheBoss:
Okay - everything should be good now. @benhoyt I fixed the bummy snap integration test by replacing the `lxd` snap being manipulated within the `test_snap_refresh` integration test with `hello-world` instead. That seemed to do the trick _at least for me locally_. @jnsgruk let me know if there are any other modifications necessary. Otherwise, I think this charm lib is ready to be merged :tada:

jnsgruk (approved):
LGTM!

benhoyt:
Thanks for the test fix for `test_snap_refresh` @NucciTheBoss!

Issue #107, opened 2023-08-23
Snap `get` and `set` methods should adhere to Python's built-in types


chanchiwai-ray:
Currently, `Snap.get` always returns the value as a string, and `Snap.set` always set the value as a string. It will be very helpful for these two methods to respect the types. For example, it will be nice that we can do the following

```python
s = snap.SnapCache()["some-snap"]
s.set({"foo": True,  "bar": {"a": 1, "b": 2.0, "c": None, "d": ["x", "y", "z"]}, "bar.greeting": "hello world", "buzz": 123, "fuzz": 123.123, "fizz": None})

x = s.get("foo")
y = s.get("bar")
y_1 = s.get("bar.a")
y_2 = s.get("bar.b")
# y_3 = s.get("bar.c") # None is not set
y_4 = s.get("bar.d")
y_5 = s.get("bar.greeting")
a = s.get("buzz")
b = s.get("fuzz")
# c = s.get("fizz") # None is not set

print(x, type(x))
print(y, type(y))
print(y_1, type(y_1))
print(y_2, type(y_2))
# print(y_3, type(y_3)) 
print(y_4, type(y_4))
print(y_5, type(y_5))
print(a, type(a))
print(b, type(b))
# print(c, type(c))
```

should prints

```shell
True <class 'bool'>
{'a': 1, 'b': 2.0, 'd': ['x', 'y', 'z'], 'greeting': 'hello world'} <class 'dict'>
1 <class 'int'>
2.0 <class 'float'>
['x', 'y', 'z'] <class 'list'>
hello world <class 'str'>
123 <class 'int'>
123.123 <class 'float'>
```

Example usages:

```python
# In app_1.py
s = snap.SnapCache()["some-snap"]
s.set({"logging.enable": True})

...

# In app_2.py
s = snap.SnapCache()["some-snap"]
# We can do 
if s.get("logging.enable"):
    do_something()
# Instead of
if s.get("logging.enable") == "True":
    do_something()
```

benhoyt:
Thanks @chanchiwai-ray. I think this might be able to be achieved with `snap set -t` and `snap get -d`. However, would you like to make a concrete proposal for the changes here, either as a PR or a discussion here? Assuming we want to be backwards compatible, I think we'd either have to 1) keep using the `get`/`set` methods but add a `json=True` keyword arg, or 2) add new methods like `set_json` / `get_json`.

chanchiwai-ray:
[...]

PR #109, opened 2023-08-24
Allow `Snap.get` and `Snap.set` to respect the Python's built-in types

chanchiwai-ray:
This patches added a `use_json` optional arguments to `Snap.get` and `Snap.set` so that the two methods can make use of the built-in types in Python while setting / getting snap configurations. 

The new option allows snap config options to be compared more natively, for example:

```python
s = snap.SnapCache()["some-snap"]
s.set({"foo": True}, use_json=True)

assert s.get("foo", use_json=True) # instead of == "True"
```

Although the required argument for the two methods did not change, the behaviour still differs a bit when setting `use_json=True`. Therefore,  the default `use_json` is set to `False` for backward compatibility. 

Fixes: #107

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -310,11 +310,11 @@ class Snap(object):
         except CalledProcessError as e:
             raise SnapError("Could not {} for snap [{}]: {}".format(args, self._name, e.stderr))
 
-    def get(self, key: str = "", *, typed: bool = False) -> Any:
+    def get(self, key: Optional[str], *, typed: bool = False) -> Any:
         """Fetch snap configuration values.
 
         Args:
-            key: the key to retrieve. Default to retrieve all values.
+            key: the key to retrieve. Default to retrieve all values for typed=True.
             typed: set to True to retrieve typed values (set with typed=True).
                 Default is to return a string.
         """
@@ -323,6 +323,9 @@ class Snap(object):
             if key:
                 return config.get(key)
             return config
+
+        if not key:
+            raise TypeError("Key must be provided when typed=False")
 
         return self._snap("get", [key]).strip()
 
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -332,7 +332,7 @@ class Snap(object):
         """
         args = []
         for key, val in configs.items():
-            value = json.dumps(val) if use_json else "'{}'".format(val)
+            value = json.dumps(val) if use_json else '"{}"'.format(val)
             args.append("{}={}".format(key, value))
 
         return self._snap("set", args)
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -310,23 +310,32 @@ class Snap(object):
         except CalledProcessError as e:
             raise SnapError("Could not {} for snap [{}]: {}".format(args, self._name, e.stderr))
 
-    def get(self, key) -> str:
+    def get(self, key: str, use_json: bool = False) -> Any:
         """Fetch a snap configuration value.
 
         Args:
             key: the key to retrieve
+            use_json: (optional) retrieve values in JSON format. Default `False`
         """
-        return self._snap("get", [key]).strip()
+        if not use_json:
+            return self._snap("get", [key]).strip()
 
-    def set(self, config: Dict) -> str:
+        configs = json.loads(self._snap("get", ["-d", key]))
+        return configs.get(key)
+
+    def set(self, configs: Dict[str, Any], use_json: bool = False) -> str:
         """Set a snap configuration value.
 
         Args:
-           config: a dictionary containing keys and values specifying the config to set.
+           configs: a dictionary containing keys and values specifying the config to set.
+           use_json: (optional) convert all values in the configs into JSON formatted string. Default `False`
         """
-        args = ['{}="{}"'.format(key, val) for key, val in config.items()]
+        args = []
+        for key, val in configs.items():
+            value = json.dumps(val) if use_json else "'{}'".format(val)
+            args.append("{}={}".format(key, value))
 
-        return self._snap("set", [*args])
+        return self._snap("set", args)
 
     def unset(self, key) -> str:
         """Unset a snap configuration value.
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 1
+LIBPATCH = 2
 
 
 # Regex to locate 7-bit C1 ANSI sequences
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -310,32 +310,35 @@ class Snap(object):
         except CalledProcessError as e:
             raise SnapError("Could not {} for snap [{}]: {}".format(args, self._name, e.stderr))
 
-    def get(self, key: str, use_json: bool = False) -> Any:
-        """Fetch a snap configuration value.
+    def get(self, key: str = "", *, typed: bool = False) -> Any:
+        """Fetch snap configuration values.
 
         Args:
-            key: the key to retrieve
-            use_json: (optional) retrieve values in JSON format. Default `False`
+            key: the key to retrieve. Default to retrieve all values.
+            typed: set to True to retrieve typed values (set with typed=True).
+                Default is to return a string.
         """
-        if not use_json:
-            return self._snap("get", [key]).strip()
+        if typed:
+            config = json.loads(self._snap("get", ["-d", key]))
+            if key:
+                return config.get(key)
+            return config
 
-        configs = json.loads(self._snap("get", ["-d", key]))
-        return configs.get(key)
+        return self._snap("get", [key]).strip()
 
-    def set(self, configs: Dict[str, Any], use_json: bool = False) -> str:
+    def set(self, config: Dict[str, Any], *, typed: bool = False) -> str:
         """Set a snap configuration value.
 
         Args:
-           configs: a dictionary containing keys and values specifying the config to set.
-           use_json: (optional) convert all values in the configs into JSON formatted string. Default `False`
+           config: a dictionary containing keys and values specifying the config to set.
+           typed: set to True to convert all values in the config into typed values while
+                configuring the snap (set with typed=True). Default is not to convert.
         """
-        args = []
-        for key, val in configs.items():
-            value = json.dumps(val) if use_json else '"{}"'.format(val)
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -310,23 +310,32 @@ def _snap_daemons(
...
         """
-        return self._snap("get", [key]).strip()
+        if not use_json:
+            return self._snap("get", [key]).strip()
 
-    def set(self, config: Dict) -> str:
+        configs = json.loads(self._snap("get", ["-d", key]))
+        return configs.get(key)
+
+    def set(self, configs: Dict[str, Any], use_json: bool = False) -> str:
```
benhoyt, line 326: We can't actually change the name of the `config` parameter without breaking backwards compatibility, as some people might be calling it by keyword (it's not a positional-only argument). And similar to above, I think we should change `use_json` to `typed`. So let's change the signature to:

```python
def set(self, config: Dict[str, Any], *, typed=False) -> str:
    ...
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -310,23 +310,32 @@ def _snap_daemons(
...
+           use_json: (optional) convert all values in the configs into JSON formatted string. Default `False`
         """
-        args = ['{}="{}"'.format(key, val) for key, val in config.items()]
+        args = []
+        for key, val in configs.items():
+            value = json.dumps(val) if use_json else '"{}"'.format(val)
+            args.append("{}={}".format(key, value))
 
-        return self._snap("set", [*args])
+        return self._snap("set", args)
```
benhoyt, line 338: I think we should have two code paths here (like you did in `get`) and pass the `-t` option, to have `snap set` validate the JSON (I know it's strictly redundant, but doesn't hurt to have the server double-check our work). How about something like this?

```python
if typed:
    kv = [f'{key}={json.dumps(val)}' for key, val in config.items()]
    return self._snap("set", ["-t"] + kv)
else:
    kv = [f'{key}={val}' for key, val in config.items()]
    return self._snap("set", kv)
```

benhoyt: Note that I've removed the double quotes `"` around the non-typed values. Python's subprocess should properly-quote keys with spaces/quotes (though we should probably add an integration test of this to check).

tests/integration/test_snap.py
```diff
@@ -52,20 +52,70 @@ def test_snap_refresh():
     lxd.ensure(snap.SnapState.Latest, classic=False, channel="latest/candidate", cohort="+")
 
 
-def test_snap_set():
+def test_snap_set_and_get_with_json():
```
benhoyt, line 55: Naming: `_with_json` -> `_typed`

tests/integration/test_snap.py
```diff
@@ -52,20 +52,70 @@ def test_snap_refresh():
...
 
-    lxd.set({"ceph.external": "false", "criu.enable": "false"})
+    assert lxd.get("criu.enable", use_json=True) == "true"
+    assert lxd.get("ceph.external", use_json=True) == "false"
 
-    assert lxd.get("ceph.external") == "false"
-    assert lxd.get("criu.enable") == "false"
 
-    lxd.set({"ceph.external": "true", "criu.enable": "true"})
+def test_snap_set_and_get_without_json():
```
benhoyt, line 111: Naming: `_without_json` -> `_untyped`

tests/unit/test_snap.py
```diff
@@ -551,26 +551,36 @@ def raise_error(cmd, **kwargs):
...
+            "list": [1, 2.0, True, False],
+            "dict": {
+                "true": True,
+                "false": False,
+                "null": None,
+                "integer": 1,
+                "float": 2.0,
+                "list": [1, 2.0, True, False],
+            },
+        }
```
benhoyt, line 569: I think we're testing too much here. We don't need to test all the types (we can assume the `json` module works), just one or two. In addition, we're not testing string here, which is one of the more important types.

Perhaps `configs = {"n": 42, "s": "string", "d": {"nested": True}}`

tests/unit/test_snap.py
```diff
@@ -551,26 +551,36 @@ def raise_error(cmd, **kwargs):
...
+
+        foo.set(configs, use_json=True)
+        args = ["{}={}".format(key, json.dumps(val)) for key, val in configs.items()]
         mock_subprocess.assert_called_with(
-            ["snap", "set", "foo", 'bar="baz"'],
+            ["snap", "set", "foo", *args],
```
benhoyt, line 574: I would avoid copying the `json.dumps` and the list comprehension into the test, and (especially with the above simplification) just include the expected output verbatim. Something like:

```python
mock_subprocess.assert_called_with(
            ["snap", "set", "foo", "-t", 'n=42', 's="string"', 'd={"nested": True}'],
)
```

tests/unit/test_snap.py
```diff
@@ -551,26 +551,36 @@ def raise_error(cmd, **kwargs):
...
-        except AssertionError:
-            # The ordering of this call can be unpredictable across Python versions.
-            # Unfortunately, the methods available to introspect the call list have also
-            # changed, so we do this, which is a little clunky.
-            mock_subprocess.assert_called_with(
-                ["snap", "set", "foo", 'qux="quux"', 'bar="baz"'],
-                universal_newlines=True,
-            )
+        foo.set(configs, use_json=False)
+        args = ['{}="{}"'.format(key, val) for key, val in configs.items()]
```
benhoyt, line 579: Similar comment as above.

And thanks for removing the quirky `AssertionError` logic -- we don't need that anymore, now that dicts are ordered in Python.

benhoyt (changes requested):
This is definitely in the right direction. Added some more suggestions/comments to see if we can refine this a bit further.

lib/charms/operator_libs_linux/v2/snap.py
chanchiwai-ray: Indeed, I didn't consider that people might calling it by keywords. Thanks for the suggestion

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -310,23 +310,32 @@ def _snap_daemons(
         except CalledProcessError as e:
             raise SnapError("Could not {} for snap [{}]: {}".format(args, self._name, e.stderr))
 
-    def get(self, key) -> str:
+    def get(self, key: str, use_json: bool = False) -> Any:
```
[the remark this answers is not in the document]
chanchiwai-ray: I think instead of making it optional `key: Optional[str] = None`, it will be better to give a default value of empty string `key: str = ""`. Coincidentally, it will make it return all values for both `typed=True` or `typed=False`. Though, it might be useful for `typed=False` since it might looks something like this for nested return values:

```shell
$ sudo snap get openstack-hypervisor compute
Key                          Value
compute.cpu-mode             host-model
compute.spice-proxy-address  10.6.0.134
compute.virt-type            kvm
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -310,23 +310,32 @@ def _snap_daemons(
...
         """Fetch a snap configuration value.
 
         Args:
             key: the key to retrieve
+            use_json: (optional) retrieve values in JSON format. Default `False`
```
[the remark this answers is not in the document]
chanchiwai-ray: Done

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -310,23 +310,32 @@ def _snap_daemons(
...
         """
-        return self._snap("get", [key]).strip()
+        if not use_json:
+            return self._snap("get", [key]).strip()
 
-    def set(self, config: Dict) -> str:
+        configs = json.loads(self._snap("get", ["-d", key]))
```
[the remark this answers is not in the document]
chanchiwai-ray: Sure, thanks for that

tests/integration/test_snap.py
chanchiwai-ray: Done

tests/integration/test_snap.py
chanchiwai-ray: Done

tests/unit/test_snap.py
chanchiwai-ray: Yeah, that's a bit overkilled... Updated.

tests/unit/test_snap.py
chanchiwai-ray: Sure

chanchiwai-ray:
It seems that integration test is failing for unrelated reasons

```
- Run configure hook of "lxd" snap if present (run hook "configure": 
-----
error: cannot communicate with server: Post "http://localhost/v2/snapctl": dial unix /run/snapd-snap.socket: connect: no such file or directory
error: cannot communicate with server: Post "http://localhost/v2/snapctl": dial unix /run/snapd-snap.socket: connect: no such file or directory
-----)
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -310,23 +310,35 @@ def _snap_daemons(
...
+            key: the key to retrieve. Default to retrieve all values.
+            typed: set to True to retrieve typed values (set with typed=True).
+                Default is to return a string.
         """
+        if typed:
+            config = json.loads(self._snap("get", ["-d", key]))
+            if key:
+                return config.get(key)
+            return config
+
```
benhoyt, line 326: I'm not comfortable with what this is doing when you call `get()` with `typed=False` -- the string returned is a table-formatted string (or perhaps empty with the way the code is currently written, as arg `''` will be sent?) which isn't very helpful.

So let's put this check here (in the `not typed` section):

```python
if not key:
    raise TypeError("key must be provided when typed=False")
```

Given that, I think `Optional[str]` would convey the type a bit better, but I don't have strong feelings about that.

benhoyt (changes requested):
One more minor change requested: let's not allow bare `get()` without `typed=True`.

lib/charms/operator_libs_linux/v2/snap.py
chanchiwai-ray: > I'm not comfortable with what this is doing when you call get() with typed=False -- the string returned is a table-formatted string

I think this is the current behaviour for nested config (although not called with `get()`):

```shell
root@sunbeam02:/tmp/lib/charms/operator_libs_linux/v2# snap get openstack-hypervisor compute
Key                          Value
compute.cpu-mode             host-model
compute.spice-proxy-address  10.6.0.134
compute.virt-type            kvm
```

```python>>> import snap
>>> s = snap.SnapCache()["openstack-hypervisor"]
>>> s.get("compute")
'Key                          Value\ncompute.cpu-mode             host-model\ncompute.spice-proxy-address  10.6.0.134\ncompute.virt-type
```

But I also agree that using `Optional[str]` will convey the meaning clearly. I will add the check and change it to `Optional[str]`

benhoyt: Ah, that's interesting to know regarding nested config -- not very useful, but oh well. :-)

Thanks for the `raise TypeError` addition. I think that will at least prevent mistakes in common cases.

benhoyt (approved):
Seems good to me now, thanks!

PR #111, opened 2023-09-14
Increase default SnapClient timeout to 30 seconds

neoaggelos:
### Summary

Increase default SnapClient timeout to 30 seconds

### Description

SnapClient has a default timeout of 5 seconds for all RPCs to
the snapd socket. If any operation takes more, an exception is
raised to the caller.

We have seen this causing issues in slow CI environments, where
a simple config-changed hook will attempt to check if a snap is
installed, timeout and move the unit to a blocked status (causing
the CI run to fail). This is augmented by the fact that typically
SnapCache is used instead of SnapClient, which only uses the default
value and does not provide a way to configure the timeout.

Simply increasing the timeout is enough to tackle the flakes.

### Notes

A possible alternative could be to surface the timeout option to SnapCache,
but that then would require each consumer of the library to set the timeout.

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 2
+LIBPATCH = 3
 
 
 # Regex to locate 7-bit C1 ANSI sequences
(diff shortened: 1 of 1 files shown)


benhoyt (approved):
Yep, this seems reasonable to me. However, we need to increment the `LIBPATCH` version at the top of the file. Please update that and then we can merge this.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -701,7 +701,7 @@ def __init__(
             opener: specifies an opener for unix socket, if unspecified a default is used
             base_url: base url for making requests to the snap client. Defaults to
                 http://localhost/v2/
-            timeout: timeout in seconds to use when making requests to the API. Default is 5.0s.
+            timeout: timeout in seconds to use when making requests to the API. Default is 30.0s.
```
benhoyt, line 704: It's a pity the [online lib docs](https://charmhub.io/operator-libs-linux/libraries/snap#snapclient) don't show these defaults in the `SnapClient.__init__` signature so that we have to repeat them here, but oh well.

benhoyt (approved):

Issue #113, opened 2023-12-13
Request: Option for apt lib to install  not backports version package


jneo8:
`freeipmi-tools` failed to install on focal

Relate: https://github.com/canonical/hardware-observer-operator/issues/128

## Reproduce

`install.py`

```python
apt.add_package("freeipmi-tools", update_cache=False)
```

```
$ python install.py
Traceback (most recent call last):
  File "/home/ubuntu/operator-libs-linux/./install.py", line 3, in <module>
    apt.add_package("freeipmi-tools", update_cache=False)
  File "/home/ubuntu/operator-libs-linux/lib/charms/operator_libs_linux/v0/apt.py", line 761, in add_package
    pkg, success = _add(p, version, arch)
                   ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/operator-libs-linux/lib/charms/operator_libs_linux/v0/apt.py", line 802, in _add
    pkg.ensure(state=PackageState.Present)
  File "/home/ubuntu/operator-libs-linux/lib/charms/operator_libs_linux/v0/apt.py", line 289, in ensure
    self._add()
  File "/home/ubuntu/operator-libs-linux/lib/charms/operator_libs_linux/v0/apt.py", line 261, in _add
    self._apt(
  File "/home/ubuntu/operator-libs-linux/lib/charms/operator_libs_linux/v0/apt.py", line 255, in _apt
    raise PackageError(
lib.charms.operator_libs_linux.v0.apt.PackageError: Could not install package(s) [['freeipmi-tools=1.6.9-2~bpo20.04.1']]: None
```
---

The apt lib choose backports version to install instead of focal-updates
```
$ apt-cache policy freeipmi-tools
freeipmi-tools:
  Installed: (none)
  Candidate: 1.6.4-3ubuntu1.1
  Version table:
     1.6.9-2~bpo20.04.1 100
        100 http://archive.ubuntu.com/ubuntu focal-backports/main amd64 Packages
     1.6.4-3ubuntu1.1 500
        500 http://archive.ubuntu.com/ubuntu focal-updates/main amd64 Packages
     1.6.4-3ubuntu1 500
        500 http://archive.ubuntu.com/ubuntu focal/main amd64 Packages
```

One way for us to resolve this issue is try to catch the target version from the command, but I feel more nice if apt lib can support it.

benhoyt:
[...]

PR #114, opened 2023-12-14
apt.py: use subprocess.run with check=True instead of check_call when capturing stdout and stderr

Perfect5th:
It's inadvisable to use `check_call` with `stdout=PIPE` and `stderr=PIPE`, as the child process will block if enough output is generated (see Note in https://docs.python.org/3/library/subprocess.html#subprocess.check_call)

We've seen some hanging subprocesses in the `landscape-server` charm due to this (and other, already addressed issues): https://bugs.launchpad.net/landscape/+bug/2033091

This change uses `subprocess.run` instead, with `capture_output=True` and `check=True`. Behaviour should be the same, except avoiding the aforementioned blocking issue.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -108,7 +108,7 @@ import re
 import subprocess
 from collections.abc import Mapping
 from enum import Enum
-from subprocess import PIPE, CalledProcessError, check_call, check_output
+from subprocess import PIPE, CalledProcessError, check_output
 from typing import Iterable, List, Optional, Tuple, Union
 from urllib.parse import urlparse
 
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
Thanks, LGTM!

benhoyt (changes requested):
Thanks! Changes look good to me, however, we need to bump up the `LIBPATCH` version here: https://github.com/canonical/operator-libs-linux/blob/a1b80b35514b2b1358745f04635596a81eaf7810/lib/charms/operator_libs_linux/v0/apt.py#L125

benhoyt (approved):
Excellent, thanks! `subprocess.run` is nice!

benhoyt:
Separately, I've noticed that we're using `e.output` (stdout) in the exception messages, meaning that the actual error message doesn't get included in the output here, it just says "None". See https://github.com/canonical/operator-libs-linux/issues/113. We'll fix that separately.

benhoyt:
That said, @Perfect5th, if you would like to submit a PR to use `e.stderr` in those exceptions, go ahead. :-)

(I notice that this change will mean that `e.output` is `b''` instead of `None`, but that doesn't help much.)

Perfect5th:
See #115

PR #115, opened 2023-12-15
apt.py: use CalledProcessError.stderr in exceptions instead of output/stdout

Perfect5th:
as mentioned in #114 , this patch changes some select `PackageError` exceptions to consume `CalledProcessError.stderr` instead of `CalledProcessError.output` (an alias for `stdout`).

These exceptions should now actually contain the error output of the subprocesses with non-zero exit codes.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -253,7 +253,7 @@ class DebianPackage:
             subprocess.run(_cmd, capture_output=True, check=True, env=env)
         except CalledProcessError as e:
             raise PackageError(
-                "Could not {} package(s) [{}]: {}".format(command, [*package_names], e.output)
+                "Could not {} package(s) [{}]: {}".format(command, [*package_names], e.stderr)
             ) from None
 
     def _add(self) -> None:
@@ -476,7 +476,7 @@ class DebianPackage:
             )
         except CalledProcessError as e:
             raise PackageError(
-                "Could not list packages in apt-cache: {}".format(e.output)
+                "Could not list packages in apt-cache: {}".format(e.stderr)
             ) from None
 
         pkg_groups = output.strip().split("\n\n")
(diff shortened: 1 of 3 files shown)


Perfect5th:
Are we okay with these `e.stderr` being `bytes`? Not too hard to ensure they are `str`s, instead.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -253,7 +253,7 @@ def _apt(
             subprocess.run(_cmd, capture_output=True, check=True, env=env)
         except CalledProcessError as e:
             raise PackageError(
-                "Could not {} package(s) [{}]: {}".format(command, [*package_names], e.output)
+                "Could not {} package(s) [{}]: {}".format(command, [*package_names], e.stderr)
```
benhoyt, line 256: Regarding bytes vs str, if we provide the `text=True` arg to `subprocess.run`, it'll be a str.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -476,7 +476,7 @@ def from_apt_cache(
             )
         except CalledProcessError as e:
             raise PackageError(
-                "Could not list packages in apt-cache: {}".format(e.output)
+                "Could not list packages in apt-cache: {}".format(e.stderr)
```
benhoyt, line 479: The call above already uses `universal_newlines=True` (which is now spelled `text=True`). So this one should already be str.

benhoyt (changes requested):

benhoyt:
Oh, can we add an integration test or two to ensure stderr actually appears correctly "in real life"?

lib/charms/operator_libs_linux/v0/apt.py
Perfect5th: Yup. Will do.

benhoyt (approved):
Looks good to me -- thanks for the updates!

PR #116, opened 2023-12-18
feat(apt): add candidate_version argument to add_package

jneo8:
Force to install package with candidate version which get from apt-cache policy output

Fix: #113


benhoyt:
Closing per https://github.com/canonical/operator-libs-linux/issues/113#issuecomment-2883260696

PR #117, opened 2024-02-02
feat(snap): add support for devmode confinement

gruyaume:
Snaps have [3 confinement levels](https://snapcraft.io/docs/snap-confinement): strict, classic and devmode. devmode snaps are useful during the development process.  Here we add support for installing snaps with the devmode level of confinement.

## Rationale

The first reason to add support for devmode is by symmetry with snaps/snapcraft. `devmode` is supported by snapcraft.io, why not support it here?

The second reason is pragmatic. The telco team has the SD-Core UPF charm which is only available in devmode. Adding strict confinement for this charm [won't be straightforward](https://github.com/canonical/sdcore-upf-snap/issues/2) but we don't want to go through the "classic" review process until we really had a try with strict confinement for it. Until then it will only be available in devmode and we need a way to move ahead with the machine charm.

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -491,7 +491,7 @@ class Snap(object):
         channel: Optional[str] = "",
         cohort: Optional[str] = "",
         revision: Optional[str] = None,
-        devmode: Optional[bool] = False,
+        devmode: bool = False,
         leave_cohort: Optional[bool] = False,
     ) -> None:
         """Refresh a snap.
@@ -537,7 +537,7 @@ class Snap(object):
         self,
         state: SnapState,
         classic: Optional[bool] = False,
-        devmode: Optional[bool] = False,
+        devmode: bool = False,
         channel: Optional[str] = "",
         cohort: Optional[str] = "",
         revision: Optional[str] = None,
@@ -583,7 +583,7 @@ class Snap(object):
                 self._install(channel, cohort, revision)
             else:
                 # The snap is installed, but we are changing it (e.g., switching channels).
-                self._refresh(channel, cohort, revision, devmode)
+                self._refresh(channel=channel, cohort=cohort, revision=revision, devmode=devmode)
 
         self._update_snap_apps()
         self._state = state
@@ -909,7 +909,7 @@ def add(
     state: Union[str, SnapState] = SnapState.Latest,
     channel: Optional[str] = "",
     classic: Optional[bool] = False,
-    devmode: Optional[bool] = False,
+    devmode: bool = False,
     cohort: Optional[str] = "",
     revision: Optional[str] = None,
 ) -> Union[Snap, List[Snap]]:
@@ -956,8 +956,13 @@ def remove(snap_names: Union[str, List[str]]) -> Union[Snap, List[Snap]]:
     snap_names = [snap_names] if isinstance(snap_names, str) else snap_names
     if not snap_names:
         raise TypeError("Expected at least one snap to add, received zero!")
-
-    return _wrap_snap_operations(snap_names, SnapState.Absent, "", False, False)
+    return _wrap_snap_operations(
+        snap_names=snap_names,
+        state=SnapState.Absent,
+        channel="",
+        classic=False,
+        devmode=False,
+    )
 
 
 @_cache_init
@@ -966,7 +971,7 @@ def ensure(
     state: str,
     channel: Optional[str] = "",
     classic: Optional[bool] = False,
-    devmode: Optional[bool] = False,
+    devmode: bool = False,
     cohort: Optional[str] = "",
     revision: Optional[int] = None,
 ) -> Union[Snap, List[Snap]]:
@@ -993,7 +998,15 @@ def ensure(
         channel = "latest"
 
     if state in ("present", "latest") or revision:
-        return add(snap_names, SnapState(state), channel, classic, devmode, cohort, revision)
+        return add(
+            snap_names=snap_names,
+            state=SnapState(state),
+            channel=channel,
+            classic=classic,
+            devmode=devmode,
+            cohort=cohort,
+            revision=revision,
+        )
     else:
         return remove(snap_names)
 
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 3
+LIBPATCH = 4
 
 
 # Regex to locate 7-bit C1 ANSI sequences
@@ -214,7 +214,7 @@ class Snap(object):
       - state: a `SnapState` representation of its install status
       - channel: "stable", "candidate", "beta", and "edge" are common
       - revision: a string representing the snap's revision
-      - confinement: "classic" or "strict"
+      - confinement: "classic", "strict", or "devmode"
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -489,6 +491,7 @@ def _refresh(
         channel: Optional[str] = "",
         cohort: Optional[str] = "",
         revision: Optional[str] = None,
+        devmode: Optional[bool] = False,
```
benhoyt, line 494: I don't think we need an optional here, just `devmode: bool = False` (with a default of False) is enough. None doesn't have a special meaning.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -530,6 +537,7 @@ def ensure(
         self,
         state: SnapState,
         classic: Optional[bool] = False,
+        devmode: Optional[bool] = False,
```
benhoyt, line 540: Same here, I think the type should be just `devmode: bool = False` (non-optional).

I know that makes it different from the type for `classic`. I think we could get away with changing that type annotation to just `bool` too, to match -- no one should be passing None here.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -566,7 +583,7 @@ def ensure(
                 self._install(channel, cohort, revision)
             else:
                 # The snap is installed, but we are changing it (e.g., switching channels).
-                self._refresh(channel, cohort, revision)
+                self._refresh(channel, cohort, revision, devmode)
```
benhoyt, line 586: I realize this isn't something you changed here (you just added to it), but it seems to me asking for trouble not to use keyword arguments here (accidental argument-order problems). Let's do:

```suggestion
                self._refresh(channel=channel, cohort=cohort, revision=revision, devmode=devmode)
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -892,6 +909,7 @@ def add(
     state: Union[str, SnapState] = SnapState.Latest,
     channel: Optional[str] = "",
     classic: Optional[bool] = False,
+    devmode: Optional[bool] = False,
```
benhoyt, line 912:
```suggestion
    devmode: bool = False,
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -937,7 +957,7 @@ def remove(snap_names: Union[str, List[str]]) -> Union[Snap, List[Snap]]:
     if not snap_names:
         raise TypeError("Expected at least one snap to add, received zero!")
 
-    return _wrap_snap_operations(snap_names, SnapState.Absent, "", False)
+    return _wrap_snap_operations(snap_names, SnapState.Absent, "", False, False)
```
benhoyt, line 960: Ick, another case where I think we should be using keyword arguments -- it's even more clear here with "False, False". Can we change these call sites to use keyword arguments too, while we're at it?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -946,6 +966,7 @@ def ensure(
     state: str,
     channel: Optional[str] = "",
     classic: Optional[bool] = False,
+    devmode: Optional[bool] = False,
```
benhoyt, line 969:
```suggestion
    devmode: bool = False,
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -970,7 +993,7 @@ def ensure(
         channel = "latest"
 
     if state in ("present", "latest") or revision:
-        return add(snap_names, SnapState(state), channel, classic, cohort, revision)
+        return add(snap_names, SnapState(state), channel, classic, devmode, cohort, revision)
```
benhoyt, line 996: Same comment here about keyword arguments. A call site with 7 positional arguments, yucky. :-) Again, not your fault -- I realize you're just adding to it.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -1014,13 +1043,17 @@ def _wrap_snap_operations(
 
 
 def install_local(
-    filename: str, classic: Optional[bool] = False, dangerous: Optional[bool] = False
+    filename: str,
+    classic: Optional[bool] = False,
+    devmode: Optional[bool] = False,
+    dangerous: Optional[bool] = False,
```
benhoyt, line 1049: I think these should all be changed to non-optional too.

tests/unit/test_snap.py
```diff
@@ -290,6 +290,37 @@ def test_can_run_snap_commands(self, mock_subprocess):
             ["snap", "install", "foo", "--classic", '--revision="123"'], universal_newlines=True
         )
 
+    @patch("charms.operator_libs_linux.v2.snap.subprocess.check_output")
+    def test_can_run_snap_commands_devmode(self, mock_subprocess):
```
benhoyt, line 294: I think it'd be better if this parameter were called `mock_check_output` as that's the function we're mocking, not the `subprocess` module itself.

benhoyt (changes requested):
Overall looks reasonable, thanks for the addition. A couple of style (keyword arg) comments as well as a note about not using `Optional[bool]` (your thoughts welcome).

Additionally, can we please add an integration test that exercises the `devmode=True` code path?

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: While I do agree, the scope of this PR wasn't to refactor the existing code. The other comments were addressed though.

tests/unit/test_snap.py
gruyaume: Done

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: Done

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: Done

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: Done

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: Done

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: Done

lib/charms/operator_libs_linux/v2/snap.py
gruyaume: Done

gruyaume:
@benhoyt as for the remark related to the integration tests, I could add a test that installs the sdcore-upf snap (which is only available in devmode), however I'm not convinced this is a great idea as the goal at some point will be to have it available as a strictly confined snap and remove devmode confinement (which will break this existing project). Do you have a suggestion here?

lib/charms/operator_libs_linux/v2/snap.py
benhoyt: Fair -- thanks.

benhoyt (approved):
Looks good to me now, thanks. We can always clean up those other `Optional[bool]`s later.

> @benhoyt as for the remark related to the integration tests, I could add a test that installs the sdcore-upf snap (which is only available in devmode), however I'm not convinced this is a great idea as the goal at some point will be to have it available as a strictly confined snap and remove devmode confinement (which will break this existing project). Do you have a suggestion here?

Yeah, that's fair enough. In that case I suggest you just test this code manually against a real Juju (you may have done that already) and report back with results here for the record.

benhoyt:
To speed things up, I've merged this pre-emptively -- if you find anything wonky in your manual testing, we can push up another PR.

Issue #120, opened 2024-05-06
Calling `Snap.ensure(....)` with the same `revision` shouldn't refresh the snap if the revision is already installed


addyess:
I didn't expect the `ensure(...)` method to actually cause a snap-refresh if the revision requested already matched the installed revision.  If one calls `ensure(...)` with a channel, and the revision is unchanged, no refresh is triggered.  

So, in my charm where i'm trying to set a specific revision on the snap, I had to add this `if` clause  prevent a constant snap refresh to a revision where it already was.

```python
    cache = snap_lib.SnapCache()
    which = cache[args.name]
    if which.revision != args.revision:
        log.info("Ensuring %s snap revision=%s", args.name, args.revision)
        which.ensure(**args.dict(exclude_none=True))
```

benhoyt:
@addyess Can you include a repro case? Is this just an optimisation, to avoid starting the `snap refresh` command if the Python code already determines it's already at the requested revision? Because presumably `snap refresh --revision=` itself won't do anything if the snap is already at that revision?

addyess:
@benhoyt I don't think it'd be hard to go back to something and reproduce in my charm code.  I found a event that was firing regularly into a path which called:

```python
    cache = snap_lib.SnapCache()
    which = cache[MY_SNAP]
    log.info("Ensuring %s snap revision=%s", MY_SNAP, MY_REV)
    which.ensure(state=SnapState.Present, name=MY_SNAP, revision=MY_REV)
```

Everytime the refresh occurred, snapd would restart the services provided by that even though it was already on that revision.

i think i didn't anticipate that `ensure` would actually trigger a restart of all the snap services even when the revision had not changed.

benhoyt:
So I just tested this with the `cups` snap on my machine (just for lack of something better to try), and there are two things I found:

1) When you just run `snap refresh cups` with additional arguments, it says `snap "cups" has no available updates` and doesn't do anything further (that's what I expected).
2) When you run `snap refresh cups --revision=1047`, that is, with a `--revision` specified, *even if it's the current installed revision*, it will do a full restart of the services. That is not what I expected `snap refresh` to do -- I would have thought that if the revision is the currently-installed revision, it would do nothing, but for whatever reason it doesn't have those smarts in snap itself.

Here your code is specifying the revision, so `--revision` is passed, and you get the second behaviour. :-(

[...]

PR #121, opened 2024-05-17
Prevent refresh by revision when the snap is already on that revision

addyess:
Addresses issue #120 

`snap refresh --revision=xyz` causes snapd to restart the service provided by that snap, even if the revision is already on the machine. In this limited case, we should not refresh the snap if it is already on the correct revision.

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 
 # Regex to locate 7-bit C1 ANSI sequences
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -584,13 +584,16 @@ class Snap(object):
                     "Installing snap %s, revision %s, tracking %s", self._name, revision, channel
                 )
                 self._install(channel, cohort, revision)
-            else:
+                logger.info("The snap installation completed successfully")
+            elif (revision is None) or (revision != self._revision):
                 # The snap is installed, but we are changing it (e.g., switching channels).
                 logger.info(
                     "Refreshing snap %s, revision %s, tracking %s", self._name, revision, channel
                 )
                 self._refresh(channel=channel, cohort=cohort, revision=revision, devmode=devmode)
-            logger.info("The snap installation completed successfully")
+                logger.info("The snap refresh completed successfully")
+            else:
+                logger.info("Snap refresh was unnecessary %s", self._name)
 
         self._update_snap_apps()
         self._state = state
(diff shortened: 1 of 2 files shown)


addyess:
@benhoyt -- tagging for review

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -584,13 +584,16 @@ def ensure(
                     "Installing snap %s, revision %s, tracking %s", self._name, revision, channel
                 )
                 self._install(channel, cohort, revision)
-            else:
+                logger.info("The snap installation completed successfully")
+            elif (revision is None) or (revision != self._revision):
```
benhoyt, line 588: In my experience using the extra parens is atypical here (and unnecessary):

```suggestion
            elif revision is None or revision != self._revision:
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -584,13 +584,16 @@ def ensure(
...
                 # The snap is installed, but we are changing it (e.g., switching channels).
                 logger.info(
                     "Refreshing snap %s, revision %s, tracking %s", self._name, revision, channel
                 )
                 self._refresh(channel=channel, cohort=cohort, revision=revision, devmode=devmode)
-            logger.info("The snap installation completed successfully")
+                logger.info("The snap refresh completed successfully")
+            else:
+                logger.info("Snap refresh was unnecessary %s", self._name)
```
benhoyt, line 596: I think this log message won't be as readable as it could be, as the snap name is just tacked on the end. Can we set it off in some way, or perhaps reorder like so?

```suggestion
                logger.info("Refresh of snap %s was unnecessary", self._name)
```

tests/unit/test_snap.py
```diff
@@ -543,6 +543,27 @@ def test_cohort(self, mock_subprocess):
             universal_newlines=True,
         )
 
+    @patch("charms.operator_libs_linux.v2.snap.subprocess")
+    def test_revision_doesnt_refresh(self, mock_subprocess):
+        mock_subprocess.check_output = MagicMock()
```
benhoyt, line 548: Would it be simpler just to mock `subprocess.check_output` directly? Then it would automatically be a `MagickMock`, right?

benhoyt (changes requested):
Couple of minor nit suggestions. But we also need to bump up the `LIBPATCH` version at the top of this file.

benhoyt (approved):

tests/unit/test_snap.py
```diff
@@ -543,6 +543,25 @@ def test_cohort(self, mock_subprocess):
...
+                "install",
+                "curl",
+                '--revision="233"',
+                '--cohort="+"',
+            ],
+            universal_newlines=True,
+        )
+
+        mock_check_output.reset_mock()
+        # ensure that calling refresh with the same revision doesn't subprocess out
```
tonyandrewmeyer, line 561:
```suggestion
        # Ensure that calling refresh with the same revision doesn't subprocess out.
```

tonyandrewmeyer (approved):

benhoyt:
Opened https://github.com/canonical/operator-libs-linux/issues/123 to track fixing the integration test.

PR #122, opened 2024-05-24
Adjustments to test with python 3.12

addyess:
While testing with python3.12, i had to bump `pyfakefs==5.5.0` due to some changes in `pathlib`

After this adjustment, i found bad tests in `test_sysctl`. 

instead of 

```python
assert my_mock.called_with(...)
```

swapped to 

```python
my_mock.assert_called_once_with(...)
```

Also, it seems codespell now doesn't like `assertIn` so let's start ignoring reasonable words

### DIFF
--- a/tests/unit/test_sysctl.py
+++ b/tests/unit/test_sysctl.py
@@ -208,8 +208,9 @@ class TestSysctlConfig(unittest.TestCase):
         config._desired_config = {"vm.swappiness": "0", "other_value": "10"}
         snapshot = config._create_snapshot()
 
-        assert mock_output.called_with(["sysctl", "vm.swappiness", "-n"])
-        assert mock_output.called_with(["sysctl", "other_value", "-n"])
+        mock_output.assert_called_once_with(
+            ["sysctl", "-n", "vm.swappiness", "other_value"], stderr=-2, universal_newlines=True
+        )
         assert snapshot == {"vm.swappiness": "1", "other_value": "5"}
 
     @patch("charms.operator_libs_linux.v0.sysctl.check_output")
@@ -222,7 +223,9 @@ class TestSysctlConfig(unittest.TestCase):
         snapshot = {"vm.swappiness": "1", "other_value": "5"}
         config._restore_snapshot(snapshot)
 
-        assert mock_output.called_with(["sysctl", "vm.swappiness=1", "other_value=5"])
+        mock_output.assert_called_once_with(
+            ["sysctl", "vm.swappiness=1", "other_value=5"], stderr=-2, universal_newlines=True
+        )
 
     @patch("charms.operator_libs_linux.v0.sysctl.check_output")
     @patch("charms.operator_libs_linux.v0.sysctl.Config._load_data")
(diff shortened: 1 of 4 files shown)


addyess:
@benhoyt I happened to find these unit tests failures when running the unit tests on python 3.12

tests/unit/test_sysctl.py
```diff
@@ -208,8 +208,9 @@ def test_create_snapshot(self, mock_load, mock_output):
         config._desired_config = {"vm.swappiness": "0", "other_value": "10"}
         snapshot = config._create_snapshot()
 
-        assert mock_output.called_with(["sysctl", "vm.swappiness", "-n"])
-        assert mock_output.called_with(["sysctl", "other_value", "-n"])
+        mock_output.assert_called_once_with(
+            ["sysctl", "-n", "vm.swappiness", "other_value"], stderr=-2, universal_newlines=True
+        )
```
dimaqq, line 213: The old test was just bad, wasn't it?
Meaning that the 2 settings were fetched serially before, and in a single command now. The code change must've been done without correcting the tests at some point prior, mustn't it?

I wonder if `stderr=-2` is actually significant, but then while the value could be checked against `ANY`, the presence of the kwarg cannot be easily ignored, which is a bit of an impasse.

Happy to have this change regardless :)

dimaqq (approved):
+1 on pyfakefs update. I vaguely remember having to do that in the past in some other repo.

addyess:
@dimaqq i don't have a clue why this integration test is failing.  It's failing in my other PR (#121) as well

benhoyt (approved):
Thanks for these changes. Mind taking a look the failing integration test, too?

dimaqq:
`
charms.operator_libs_linux.v2.snap.SnapError: Snap: 'lxd'; command ['snap', 'refresh', 'lxd', '--channel="latest"'] failed with output = '2024-05-24T03:43:44Z INFO Waiting for "snap.lxd.daemon.service" to stop.\n'
`

Perhaps the old version needs more time to stop?
Or maybe we're not reading the entire output?

addyess:
> Thanks for these changes. Mind taking a look the failing integration test, too?

Aye:

looks like refreshing the lxd snap fails if we're in a big hurry?
```
2024-05-24T03:43:44Z INFO Waiting for "snap.lxd.daemon.service" to stop.\n
```

addyess:
there's also this in the pytest logs:

```
tests/integration/test_snap.py::test_snap_set_and_get_with_typed 
-------------------------------- live log call ---------------------------------
INFO     charms.operator_libs_linux.v2.snap:snap.py:589 Refreshing snap lxd, revision None, tracking latest
- Run configure hook of "lxd" snap if present (run hook "configure": error: cannot communicate with server: Post "http://localhost/v2/snapctl": dial unix /run/snapd-snap.socket: connect: no such file or directory)
FAILED
```

looks like something has gone wrong with snapd on the machine during refresh that it cannot communicate with lxd??

benhoyt:
Ah yes, I just noticed that too @addyess. Weird, it's like the snap daemon isn't running or something. We'll let you off the hook. ;-) I'll merge this PR and we can address the integration test failures separately.

benhoyt:
Opened https://github.com/canonical/operator-libs-linux/issues/123 to track fixing the integration test.

addyess:
>  We'll let you off the hook. ;-)

This...time...only...

PR #125, opened 2024-05-29
snap.py: fix snap.get optional key

### LINKED ISSUE #124: Snap get None key fails with `TypeError: expected str, bytes or os.PathLike object, not NoneType`
get's interface allows for key to be optional when typed is provided:
https://github.com/canonical/operator-libs-linux/blob/522d54e43655830fdb99589721e8a2f1af2a743a/lib/charms/operator_libs_linux/v2/snap.py#L313

But, if you actually provide None, you get a TypeError:

```python
snap.SnapCache()["openstack-hypervisor"].get(None, typed=True)
*** TypeError: expected str, bytes or os.PathLike object, not NoneType
```

The key is passed as is to the subprocess call, but subprocess does not support null arguments:
```python3
subprocess.check_output(["cat", None], universal_newlines=True)
*** TypeError: expected str, bytes or os.PathLike object, not NoneType
```

(tested with jammy's python 3.10.12)

gboutry:
Fix snap.get method to handle the case where the key is falsy. Treat it as no key provided, which results in the expected behavior to get all values from snap configs.

Fix #124

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -83,7 +83,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 6
+LIBPATCH = 7
 
 
 # Regex to locate 7-bit C1 ANSI sequences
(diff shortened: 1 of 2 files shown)


tonyandrewmeyer (approved):
Thanks for this!

benhoyt:
Thanks!

Issue #123, opened 2024-05-24
Fix failing test_snap_set_and_get_with_typed integration test


benhoyt:
This is failing with an error connecting to the snapd socket: 

```
tests/integration/test_snap.py::test_snap_set_and_get_with_typed 
-------------------------------- live log call ---------------------------------
INFO     charms.operator_libs_linux.v2.snap:snap.py:590 Refreshing snap lxd, revision None, tracking latest
error: cannot perform the following tasks:
- Run configure hook of "lxd" snap if present (run hook "configure": error: cannot communicate with server: \
    Post "http://localhost/v2/snapctl": dial unix /run/snapd-snap.socket: connect: no such file or directory)
FAILED
```

Other similar tests seem to be succeeding.

PR #128, opened 2024-07-16
feat: enable juju-systemd-notices to observe snap services

### LINKED ISSUE #126: juju-systemd-notices does not support nested service names like snap.slurm.slurmd
The `juju-systemd-notices` charm library does not support observing snap service state. 

The issue originates from the fact that snap service names are "nested," where the service name is interpolated with `.`. These snap-specific service names cause the charm library to bork because when you go to define a custom event for the watched service in `ops`, the charm will fail to define the event as `.` is an invalid character for event names. Screenshot below shows the error encountered when passing the service name as `snap.slurm.slurmd`:

image

NucciTheBoss:
Fixes #126 

This pull request enables support for observing snap services using the `juju-systemd-notices` charm library.

The key change in this pull request is modifying how the notices daemon configures how it observes services on the machine. Rather than configure which services to observe by reading the `hooks` directory, the daemon instead reads a _watch.yaml_ file in the charm directory. Here's what the _watch.yaml_ file would look like in a deployed charm using the notices daemon:

```yaml
---
snap.slurm.slurmd: slurmd
snap.slurm.munged: munged
```

Originally, `juju-systemd-notices` used a regex that parsed hook names to determine which services to observe with the notices daemon. This approach worked until we needed to start working with nested services - e.g. have `.` in the service name like `snap.slurm.slurmd` - so we had to find a new way to configure how to observe services, but also ensure that the correct hook would be fired by the charm. We decided to use a key value approach where the key is the service name to observe over DBus, and the value is the valid event name alias used to trigger the start/stop events.

Now, when the notices daemon captures the event, it uses the configured alias to determine which hook to fire, but can still watch DBus using the correct service name.

### Misc.

1. I tried using systemd unit aliases, but that didn't work. DBus will "follow the symlink" to the original service name - e.g. `snap.slurm.slurmd` - and not the unit alias `slurmd`.
2. We had to update _tox.ini_ because `ruff` now requires the subcommand `check` to perform linting actions.
3. We likely need to bump the LIBAPI version since the `SystemdNotices` __init__ constructor now takes a variable amount of `Service` objects rather than a `List[str]`. We used a `Service` dataclass since it made it easy to provide service name aliases and set a default alias if none was provided by the charm author.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
+++ b/lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
@@ -42,7 +42,7 @@ class ApplicationCharm(CharmBase):
         super().__init__(*args, **kwargs)
 
         # Register services with charm. This adds the events to observe.
-        self._systemd_notices = SystemdNotices(self, Service("snap.slurm.slurmd", alias="slurmd"))
+        self._systemd_notices = SystemdNotices(self, [Service("snap.slurm.slurmd", alias="slurmd")])
         self.framework.observe(self.on.install, self._on_install)
         self.framework.observe(self.on.stop, self._on_stop)
         self.framework.observe(self.on.service_slurmd_started, self._on_slurmd_started)
--- a/lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
+++ b/lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
@@ -91,9 +91,8 @@ import subprocess
 import textwrap
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Mapping, Optional
+from typing import List, Optional, Union
 
-import yaml
 from dbus_fast.aio import MessageBus
 from dbus_fast.constants import BusType, MessageType
 from dbus_fast.errors import DBusError
@@ -113,10 +112,11 @@ LIBPATCH = 1
 
 # juju-systemd-notices charm library dependencies.
 # Charm library dependencies are installed when the consuming charm is packed.
-PYDEPS = ["dbus-fast>=1.90.2", "pyyaml>=6.0.1"]
+PYDEPS = ["dbus-fast>=1.90.2"]
 
 _logger = logging.getLogger(__name__)
 _juju_unit = None
+_observed_services = {}
 _service_states = {}
 _DBUS_CHAR_MAPPINGS = {
     "_5f": "_",  # _ must be first since char mappings contain _.
@@ -161,9 +161,6 @@ class Service:
     name: str
     alias: Optional[str] = None
 
-    def __post_init__(self) -> None:  # noqa D105
-        self.alias = self.alias or self.name
-
 
 class ServiceStartedEvent(EventBase):
     """Event emitted when service has started."""
@@ -176,24 +173,25 @@ class ServiceStoppedEvent(EventBase):
 class SystemdNotices:
     """Observe systemd services on your machine base."""
 
-    def __init__(self, charm: CharmBase, *services: Service) -> None:
+    def __init__(self, charm: CharmBase, services: List[Union[str, Service]]) -> None:
         """Instantiate systemd notices service."""
         self._charm = charm
-        self._services = services
+        self._services = [Service(s) if isinstance(s, str) else s for s in services]
         unit_name = self._charm.unit.name.replace("/", "-")
         self._service_file = Path(f"/etc/systemd/system/juju-{unit_name}-systemd-notices.service")
 
         _logger.debug(
             "Attaching systemd notice events to charm %s", self._charm.__class__.__name__
         )
-        for service in self._services:
-            self._charm.on.define_event(f"service_{service.alias}_started", ServiceStartedEvent)
-            self._charm.on.define_event(f"service_{service.alias}_stopped", ServiceStoppedEvent)
+        for s in self._services:
+            event = s.alias or s.name
+            self._charm.on.define_event(f"service_{event}_started", ServiceStartedEvent)
+            self._charm.on.define_event(f"service_{event}_stopped", ServiceStoppedEvent)
 
     def subscribe(self) -> None:
         """Subscribe charmed operator to observe status of systemd services."""
         self._generate_hooks()
-        self._generate_config()
+        self._generate_service()
         self._start()
 
     def stop(self) -> None:
@@ -205,53 +203,55 @@ class SystemdNotices:
     def _generate_hooks(self) -> None:
         """Generate legacy event hooks for observed systemd services."""
         _logger.debug("Generating systemd notice hooks for %s", self._services)
-        start_hooks = [Path(f"hooks/service-{s.alias}-started") for s in self._services]
-        stop_hooks = [Path(f"hooks/service-{s.alias}-stopped") for s in self._services]
+        events = [s.alias or s.name for s in self._services]
+        start_hooks = [Path(f"hooks/service-{e}-started") for e in events]
+        stop_hooks = [Path(f"hooks/service-{e}-stopped") for e in events]
         for hook in start_hooks + stop_hooks:
             if hook.exists():
                 _logger.debug("Hook %s already exists. Skipping...", hook.name)
             else:
                 hook.symlink_to(self._charm.framework.charm_dir / "dispatch")
 
-    def _generate_config(self) -> None:
-        """Generate watch file for systemd notices daemon."""
-        _logger.debug("Generating watch file for %s", self._services)
-        config = {"services": {s.name: s.alias for s in self._services}}
+    def _generate_service(self) -> None:
+        """Generate systemd service file for notices daemon."""
+        _logger.debug("Generating service file %s", self._service_file.name)
+        if self._service_file.exists():
+            _logger.debug("Overwriting existing service file %s", self._service_file.name)
 
-        config_file = self._charm.framework.charm_dir / "watch.yaml"
-        if config_file.exists():
-            _logger.debug("Overwriting existing watch file %s", config_file.name)
-        config_file.write_text(yaml.dump(config))
-        config_file.chmod(0o600)
+        services = [f"{s.name}={s.alias or s.name}" for s in self._services]
+        self._service_file.write_text(
+            textwrap.dedent(
+                f"""
+                [Unit]
+                Description=Juju systemd notices daemon
+                After=multi-user.target
+
+                [Service]
+                Type=simple
+                Restart=always
+                WorkingDirectory={self._charm.framework.charm_dir}
+                Environment="PYTHONPATH={self._charm.framework.charm_dir / "venv"}"
+                ExecStart=/usr/bin/python3 {__file__} --unit {self._charm.unit.name} {' '.join(services)}
+
+                [Install]
+                WantedBy=multi-user.target
+                """
+            ).strip()
+        )
+
+        _logger.debug(
+            "Service file %s written. Reloading systemd manager configuration",
+            self._service_file.name,
+        )
 
     def _start(self) -> None:
         """Start systemd notices daemon to observe subscribed services."""
         _logger.debug("Starting %s daemon", self._service_file.name)
-        if self._service_file.exists():
-            _logger.debug("Overwriting existing service file %s", self._service_file.name)
-        self._service_file.write_text(
-            textwrap.dedent(
-                f"""
-                    [Unit]
-                    Description=Juju systemd notices daemon
-                    After=multi-user.target
 
-                    [Service]
-                    Type=simple
-                    Restart=always
-                    WorkingDirectory={self._charm.framework.charm_dir}
-                    Environment="PYTHONPATH={self._charm.framework.charm_dir / "venv"}"
-                    ExecStart=/usr/bin/python3 {__file__} {self._charm.unit.name}
-
-                    [Install]
-                    WantedBy=multi-user.target
-                """
-            ).strip()
-        )
-        _logger.debug("Service file %s written. Reloading systemd", self._service_file.name)
+        # Reload systemd manager configuration so that it will pick up notices daemon.
         _daemon_reload()
-        # Notices daemon is enabled so that the service will start even after machine reboot.
-        # This functionality is needed in the event that a charm is rebooted to apply updates.
+
+        # Enable notices daemon to start after machine reboots.
         _enable_service(self._service_file.name)
         _start_service(self._service_file.name)
         _logger.debug("Started %s daemon", self._service_file.name)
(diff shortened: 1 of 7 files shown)


lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -111,12 +113,11 @@ def _on_slurmd_stopped(self, _: ServiceStoppedEvent) -> None:
 
 # juju-systemd-notices charm library dependencies.
 # Charm library dependencies are installed when the consuming charm is packed.
-PYDEPS = ["dbus-fast>=1.90.2"]
+PYDEPS = ["dbus-fast>=1.90.2", "pyyaml>=6.0.1"]
```
benhoyt, line 116: Couple of minor things here: `PyYAML` seems to be the canonical spelling/case of the [project name on PyPI](https://pypi.org/project/PyYAML/). Would `PyYAML==6.*` be better, so that it doesn't pull in a potentially-breaking 7.x version?

benhoyt: Also, we should definitely update LIBPATCH. Ideally we should do this change without having to update LIBAPI (that is, no breaking changes).

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -148,6 +149,22 @@ def _systemctl(*args) -> None:
...
+    Args:
+        name: Name of systemd service to observe on dbus.
+        alias: Event name alias for service.
+    """
+
+    name: str
+    alias: Optional[str] = None
+
+    def __post_init__(self) -> None:  # noqa D105
+        self.alias = self.alias or self.name
```
benhoyt, line 165: Nit: in terms of data modelling, I don't think we should do this. If it has an alias, alias should be set, otherwise None (this also makes the field type correctly optional). Code that uses `Service.alias` should do the fallback, for example `s.alias or s.name`

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -159,7 +176,7 @@ class ServiceStoppedEvent(EventBase):
 class SystemdNotices:
     """Observe systemd services on your machine base."""
 
-    def __init__(self, charm: CharmBase, services: List[str]) -> None:
+    def __init__(self, charm: CharmBase, *services: Service) -> None:
```
benhoyt, line 179: I don't think we should make a breaking change for this relatively minor improvement. It's easy in Python to handle both dynamically. A couple of possibilities:

1) Make the type `List[str | Service]` or similar.
2) Make it so if the first argument is a list, it treats it as a list of str, otherwise *args.
2) Keep it `List[str]`, and if a string has an `=` in it, use the right hand side as an alias. Something like `SystemdNotices(['snap.slurm.slurmd=slurmd', 'noalias'])`

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -256,6 +292,16 @@ def _dbus_path_to_name(path: str) -> str:
     return name
 
 
+@functools.lru_cache(maxsize=32)
```
benhoyt, line 295: Why 32? Would 1 be sufficient, given there are no arguments? (We could in future use `functools.cache`, but that's Python 3.9+ only, and this repo supports 3.8.)

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -256,6 +292,16 @@ def _dbus_path_to_name(path: str) -> str:
...
+def _read_config() -> Mapping[str, str]:
+    """Read systemd notices daemon configuration to service names and aliases."""
+    config_file = Path.cwd() / "watch.yaml"
+    _logger.debug("Loading observed services from configuration file %s", config_file)
+
+    with config_file.open("rt") as fin:
+        return yaml.safe_load(fin)["services"]
```
benhoyt, line 302: I do like the more explicit way of telling the background process about the service names (and aliases). However, writing to a YAML config seems a bit heavy (and requires an addition to the PYDEPS). What about avoiding the YAML dependency and extra complexity by using command line args:

```
services = [f'{s.name}={s.alias}' if s.alias else s.name for s in self._services]
service_file.write_text(f"""
...
ExecStart=/usr/bin/python3 {__file__} {self._charm.unit.name} {' '.join(services)}
...
""")
```

And then the main() would use argparse with nargs='*' (I think it is) and read them from the CLI argument.

It may also make testing slightly easier.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -386,7 +422,7 @@ async def _async_load_services() -> None:
 
     # Loop through all the services and be sure that a new watcher is
     # started for new ones.
-    for service in watched_services:
+    for service in watched_services.keys():
```
benhoyt, line 425: Why is `.keys()` necessary here? Iterating keys is the default for dicts.

benhoyt (commented):
Hi @NucciTheBoss -- nice work on this. A couple of comments about backwards compatibility, and a suggestion to simplify by using CLI arguments instead of `watch.yaml`.

While you're here, would you also have a few spare cycles to look at the [`test_snap_set_and_get_with_typed` integration test failure](https://github.com/canonical/operator-libs-linux/issues/123)? (In a separate PR.) If you can, that would be amazing.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I did think about using `functools.cache`, but yeah, it wasn't added until Python 3.9+. I mostly used `32` because it felt like a good number; I haven't used `functools.lru_cache` much but I did know that I didn't want to keep reopening the `watch.yaml` file every loop :sweat_smile:

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I like this suggestion. I'll test it out. Big thing is ensuring how we can "bubble down" the observed service configuration to the daemon's methods. Could likely use a global map similar to `_service_states` where we just set a map of the watched services when `argparse` parses the arguments passed to the daemon.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Force of habit :sweat_smile:

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I did think about going with option 1 after changing the function signature to `*services: Service`. It's my preferred option compared to the other 2. 

Key thing to note is that we'll need to set the signature to `List[Union[str, Service]]`. Support for using `|` to represent a union type annotation [wasn't added until Python 3.10+](https://docs.python.org/3/library/typing.html#typing.Union).

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: No longer using `functools.lru_cache` as of https://github.com/canonical/operator-libs-linux/pull/128/commits/c8f0767563b3e317b91c0406c0872db930a09830.

Now instead using a global map to track observed services and aliases.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I went ahead with this approach. Works nicely!

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Reverted to compatible API so we only have to bump the LIBPATCH version.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Dropped `PyYAML` as a dependency in https://github.com/canonical/operator-libs-linux/pull/128/commits/c8f0767563b3e317b91c0406c0872db930a09830

NucciTheBoss:
@benhoyt thank you for the feedback! I have applied the latest updates to this branch. The pull request is ready for review again :smiley: 

Re. https://github.com/canonical/operator-libs-linux/issues/123 - seems like an intermittent issue. I could potentially look at it if I have some extra time this pulse :watch::hourglass:

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -310,8 +346,10 @@ async def _send_juju_notification(service: str, state: str) -> None:
     if service.endswith(".service"):
         service = service[0:-len(".service")]  # fmt: skip
 
+    global _observed_services
```
benhoyt, line 349: You actually only need the `global` statement when you're re-assigning the global; here you're just referencing it (you've already assigned the name at the global level), so no need for `global`.

I'd prefer if this was more explicitly passed down, rather than a global, in any case. Is that relatively easy to achieve? (Similar with `_juju_unit`.)

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -115,8 +116,8 @@ def _on_slurmd_stopped(self, _: ServiceStoppedEvent) -> None:
 
 _logger = logging.getLogger(__name__)
 _juju_unit = None
+_observed_services = {}
```
benhoyt, line 119: Would `_service_aliases` be more apt?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -363,30 +401,13 @@ async def _async_load_services() -> None:
...
-    _logger.info("Services from hooks are %s", watched_services)
-    if not watched_services:
-        return
-
     bus = await MessageBus(bus_type=BusType.SYSTEM).connect()
 
     # Loop through all the services and be sure that a new watcher is
     # started for new ones.
-    for service in watched_services:
+    global _observed_services
```
benhoyt, line 408: Again, no `global` required; you're just referencing it, not reassigning it.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -199,27 +231,31 @@ def subscribe(self) -> None:
                 Restart=always
                 WorkingDirectory={self._charm.framework.charm_dir}
                 Environment="PYTHONPATH={self._charm.framework.charm_dir / "venv"}"
-                ExecStart=/usr/bin/python3 {__file__} {self._charm.unit.name}
+                ExecStart=/usr/bin/python3 {__file__} --unit {self._charm.unit.name} {' '.join(services)}
```
benhoyt, line 234: Why do we pass `--unit` here now, when we weren't before?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -462,13 +483,19 @@ def _main():
...
-    parser.add_argument("unit", type=str)
+    parser.add_argument("--unit", type=str)
+    parser.add_argument("services", nargs="*")
     args = parser.parse_args()
 
     # Intentionally set as global.
     global _juju_unit
     _juju_unit = args.unit
 
+    global _observed_services
```
benhoyt, line 494: Again, `global` not required. You're not assigning the name as you're only using it via `_observed_services[service] = alias`. You'd need it if you were assigning the name itself, like  `_observed_services = {}`.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/python3
-# Copyright 2023 Canonical Ltd.
+# Copyright 2023-2024 Canonical Ltd.
```
benhoyt, line 2: Let's not forget to increment `LIBPATCH`.

benhoyt (changes requested):
Looking good! I do think it's more straight-forward without the watch.yaml file. Just one more iteration: 1) required: increment LIBPATCH, and 2) optional: can we remove the globals?

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Mostly a UX preference for when debugging the notices daemon if there are issues. I try to avoid using args for two or more different objects since it might not always be clear what the order is. E.g. does the unit name come last or first after providing service names: https://clig.dev/#arguments-and-flags

I added the `--unit` flag so it's more clear that it's different from the service names and aliases that are being passed to the notices daemon.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: I think `_observed_services` is a bit more clear because this map contains both the service name and alias as opposed to just one or the other.

If `_observed_services` was just a list of the aliases, then I would agree that `_service_aliases` would be a better name. However I think `_observed_services` is fine here since it's serving as a holder for both the service names and aliases.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: `global` can be used for explicitly declaring that we're using a global variable within the scope of the function, but yes, we can reference it implicitly as well.

I prefer to use the globals here since they're effectively constants that we're setting in `_main()` before starting the DBus observer. I don't necessarily want to expand the number of arguments that the daemon's methods must accept to pass down static values. I also think it's simpler as we know that all references to `_juju_unit`, `_observed_services`, and `_service_states` are referring to the same variable rather than needing to chase down references within the daemon's functions and see what is bubbled back up.

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Dropped global reference in https://github.com/canonical/operator-libs-linux/pull/128/commits/e3ab7193bffcdc5b04efb6fcd37f4b6e960ea7fb

lib/charms/operator_libs_linux/v0/juju_systemd_notices.py
NucciTheBoss: Dropped global reference in https://github.com/canonical/operator-libs-linux/pull/128/commits/e3ab7193bffcdc5b04efb6fcd37f4b6e960ea7fb

NucciTheBoss:
Addressed latest feedback comments and bumped `LIBPATCH` to 2. I think these updates look pretty good. I decided to keep the global variables for now, but we can refactor in the future if the lib becomes to complex and the global values are frequently updated by multiple methods.

Let me know if you have any more questions or comments!

benhoyt (approved):
Great, thanks for the tweaks and LIBPATCH bump. Good with me!

PR #130, opened 2024-08-21
Fail and retry on "Some index files failed to download"

nobuto-m:
In scripting apt operations, adding `--error-on=any` to `apt-get update` is necessary to capture erros like "Some index files failed to download". Otherwise, charms move onto package installations with a stale index.

equivalent to: https://github.com/juju/charm-helpers/pull/911
ref: https://bugs.launchpad.net/ubuntu/+source/apt/+bug/1693900


benhoyt (changes requested):
Sounds reasonable to me, thanks. Can you also please update the LIBPATCH version in apt.py?

benhoyt (approved):

Issue #135, opened 2024-11-14
Noble apt `RepositoryMapping` unsupported


yanksyoon:
The RepositoryMapping object in apt lib only reads from "/etc/apt/sources.list" and "/etc/apt/sources.list.d/*.list".

Noble's sources are located at `/etc/apt/sources.list.d/ubuntu.sources`, hence raising `InvalidSourceError`.

dimaqq:
A PR would be very welcome :)

yanksyoon:
@dimaqq out of curiosity, may I ask what team is currently maintaining this repository please?

benhoyt:
Our team has picked up maintenance of this repo. We don't have a ton of cycles to spend on maintenance, but this particular one seems like a significant issue! @james-garner-canonical, getting this charm lib to work on Noble seems like a noble task, so assigning to you -- perhaps you can take a look at this in the next week or two.

NucciTheBoss:
@james-garner-canonical have you had chance to tackle this issue yet? I'm getting ganked by it as the HPC team is upgrading the Slurm charms to Noble so that we can have cool kid dkms modules and fancy new login node daemons.
 
Looking at the required changes, we'll likely need to rework how the sources lists are constructed as Noble uses the [deb822 standard](https://manpages.ubuntu.com/manpages/noble/en/man5/sources.list.5.html), which the apt charm library doesn't support; _it only supports the one-line style format_. The extensions are separated between `.list` and `.sources` respectively, so that could be used as the indicator for which parser should be used.

The parser itself seems pretty simple itself by just iterating the over sources list line by line. `deb822` might be more interesting since the sources are specified in blocks rather than lines.

james-garner-canonical:
Thanks for raising this again @NucciTheBoss, and for the helpful info about the deb822 standard. I'll look into this today

PR #137, opened 2024-11-21
feat: load deb822 sources; use add-apt-repository to add

james-garner-canonical:
Ubuntu 24.04 uses [deb288 format](https://manpages.ubuntu.com/manpages/noble/en/man5/sources.list.5.html) sources, which python-libjuju does not currently support (as pointed out by @yanksyoon and @NucciTheBoss in #135). This PR adds support for reading `*sources` files.

`RepositoryMapping` now automatically findes them in `/etc/apt/sources.list.d` when initialised. Additionally, to avoid an `InvalidSourceError` if `/etc/apt/sources.list` exists but only includes a comment about the new sources location (as it does on noble), we ignore this error if `/etc/apt/sources.list.d/ubuntu.sources` exists.

This PR also cleans up the `add` method by calling `add-apt-repository` internally instead of writing directly to a file. Note that this means that adding a disabled repository no longer has any effect (previously it would write a a file containing a single, commented-out one-line repository definition).

Original PR content below.

---

Ubuntu 24.04 uses [deb288 format](https://manpages.ubuntu.com/manpages/noble/en/man5/sources.list.5.html) sources, which python-libjuju does not currently support (as pointed out by @yanksyoon and @NucciTheBoss in #135). This PR adds support for reading `*sources` files from `/etc/apt/sources.list.d`.

While there are existing deb288 parsing implementations, such as [python-apt](https://salsa.debian.org/apt-team/python-apt/-/blob/main/aptsources/_deb822.py) and [python-debian](https://salsa.debian.org/python-debian-team/python-debian/-/blob/master/src/debian/deb822.py), this charm-lib currently has no external dependencies, nor should it have any, so I've taken a stab at implementing deb288 parsing in the `RepositoryMapping` class.

In `__init__`, if files matching `/etc/apt/sources.list.d/*sources` exist, they are loaded using the new `load_deb822` method. Loading and parsing is broken down into collecting the lines that make up a paragraph/entry (`_iter_deb822_paragraphs`), parsing the lines into key/value pairs (`_get_deb822_options`), and validating these entries to construct `DebianRepository`s (`_parse_deb822_paragraph`).

Unit tests cover this logic with text from a good deb822 format file, and a malformed version of this file. However, there are no doubt a number of edge cases that haven't been covered here.

I haven't made any changes to integration tests, but with the functionality added they currently load a `.sources` file in `tests/integration/test_apt.py`'s `test_install_package_external_repository` and `test_list_file_generation_external_repository`. However it might be good to explicitly test this in integration tests too.

I'd welcome suggestions for test cases, as well as pointing out any deb822 features that are missing or not implemented correctly.


lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1314,6 +1318,203 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+        repositories: List[DebianRepository] = []
+        errors: List[InvalidSourceError] = []
+        for paragraph in cls._iter_deb822_paragraphs(lines):
+            try:
+                repos = cls._parse_deb822_paragraph(paragraph, filename=filename)
+            except InvalidSourceError as e:
+                errors.append(e)
+            else:
+                repositories.extend(repos)
+        return repositories, errors
```
NucciTheBoss, line 1378: I know regexes invoke a lot of opinions among software engineers, but I found this regex to be pretty useful for parsing deb822 stanzas. You can see it in action on regex101 - https://regex101.com/r/UriO7l/1:

```python3
_deb822_matcher = re.compile(
    r"""
    (?:Enabled:\s*)?(?P<enabled>.+)?\s?
    Types:\s*(?P<repo_types>.{3,})\s
    URIs:\s*(?P<uris>.+)\s
    Suites:\s*(?P<suites>.+)\s
    (?:Components:\s*)?(?P<components>.+)?\s?
    (?:Signed-By:\s*)?(?P<gpg_key>.+)?\s?
    (?P<options>((.*:\s*)(.+)\s?)*)?
    """,
    re.VERBOSE,
)
```

Then you can glom all the stanzas in a `.sources` file like so:

```python3
for stanza in _deb822_matcher.finditer(content):
    groups = stanza.groupdict()
    enabled = groups.pop("enabled")
    repo_types = groups.pop("repo_types").split()
    uris = groups.pop("uris").split()
    suites = groups.pop("suites").split()
    components = groups.pop("components").split()
    gpg_key = groups.pop("gpg_key")
    raw_options = groups.pop("options")

    if enabled not in ["yes", "no", None]:
        raise InvalidSourceError("...")
    else:
        enabled = True if enabled == "yes" or enabled is None else False

    if len(suites) == 1 and suites[0].endswith("/") and components is not None:
        raise InvalidSourceError("...")
    elif components is None:
        raise InvalidSourceError("...")

    options = {}
    for option in raw_options.splitlines():
        k, v = option.split(":", maxsplit=1)
        options[k] = v.strip()
```

You do lose some of the granularity you're getting such as the line numbers that you're parsing, but there is the added benefit of not needing to maintain a custom file parser. _Trade one form of complexity for another_. The one benefit of the deb822 format is that it's easier for both humans to understand and for machines to manipulate :sweat_smile:

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1314,6 +1318,203 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+            `load_deb822` strips out comments entirely when parsing a file into paragraphs, and
+                assumes that comments have been removed when parsing individual paragraphs/entry,
+                instead only reading the 'Enabled' key to determine if an entry is enabled
+        """
+        with open(filename, "r") as f:
+            repos, errors = self._parse_deb822_lines(f, filename=filename)
+
+        for repo in repos:
+            repo_identifier = "{}-{}-{}".format(repo.repotype, repo.uri, repo.release)
+            self._repository_map[repo_identifier] = repo
```
NucciTheBoss, line 1348: I wonder if there should be a new class - e.g. something like `Debian822Repository` - or a new attribute on `DebianRepository` - e.g. `DebianRepository(...).source_format` - for sources that are written in deb822 style format :thinking: 

Reason saying is that I think the current `DebianRepository` class is well suited for one-line style sources, but not deb822 style. If a particular `DebianRepository` generated from the deb822 source, how does that get mapped back correctly into the `*.sources` file? Even though we're able to convert the deb822 style source to multiple one-lines, the `RepositoryMapping` cache loses how the source is actually represented in the `*.sources` file it was read from. Perhaps if we had some tag or differentiation between the two formats on `DebianRepository`, we can signal that the source should be treated different when dumping a source into a file under `/etc/apt/sources.list.d/` or modifying an existing source in a `*.source` file.

NucciTheBoss (changes requested):
Looking good! :star_struck: 

 I just have a couple comments/suggestions around how deb822 style sources are being handled in tandem with one-line style sources.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1314,6 +1318,203 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+                enabled=enabled,
+                repotype=repotype,
+                uri=uri,
+                release=suite,
+                groups=components,
+                filename=filename,
+                gpg_key_filename=gpg_key,  # TODO: gpg_key can be a literal key, not just a filename
+                options=options,
+            )
+            for repotype, uri, suite in itertools.product(repotypes, uris, suites)
```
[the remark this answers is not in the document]
james-garner-canonical: This is a great point, I didn't look into how `DebianRepository` objects are used or the methods they support -- like disabling a source. The repository object we end up with will have to be aware of the format. My initial inclination is a `Debian822Repository` inheriting from `DebianRepository` or at least implementing the same 'protocol'.

This also makes me wonder if we do need to parse in commented out deb822 style entries and treat them as disabled, which could be a pain. More on parsing in next reply.

lib/charms/operator_libs_linux/v0/apt.py
james-garner-canonical: Note in case you read the previous comment in emails, I've edited it a bit

lib/charms/operator_libs_linux/v0/apt.py
james-garner-canonical: I think this is actually a good case for inheritance for maximum backwards compatibility. A `Debian822Repository` is a `DebianRepository`, just with different implementation details of some methods.

This is perhaps a good point to think about how to handle a `Signed-by` key that was provided inline. It's not clear to me how best to handle the existing `gpg_key` property in this case, which is expected to be a string filename. There doesn't seem to be a property for getting the key file contents directly -- the other key related methods are all about adding a key to a repository.

 Here are some options for how we could handle the inline key case.
1. return an empty string (default value), and document that users should check `gpg_key_inline` in this case
    - old code won't instantly break
    - old code will silently assume that inline key entries aren't signed
        - maybe ok if they raise an error in this case, since they can then be fixed (but bad to break them)
        - bad if instead you get a silent behaviour change (seems more likely to me, but idk)
    - new code will have to do `isinstance` or `hasattr` or catch `AttributeError` when checking for `gpg_key_inline`
3. return a new sentinel value to indicate that you should check `gpg_key_inline`
    - instantly breaks old code expecting a string
    - doesn't hide any errors with treating the entry as unsigned
    - new code has to switch based on the return value (but probably is already to handle the empty string case indicating no key)
5. just return the inline key instead of a filename
    - instantly breaks old code expecting a filename
    - doesn't hide any errors with treating the entry as unsigned
    - annoying for all users to check if it's a filename or key
7. try to do something clever by writing the inline key to a temporary file (in memory?) and returning the path to that
    - old code works perfectly
    - added complexity? what could go wrong here?

james-garner-canonical:
TL;DR: a deb822 source can specify multiple one-line-style repositories. This PR addressed this by parsing into multiple `DebianRepository` objects, but this is problematic because such objects provide a `disable` method to disable a repository. It's not clear if there's a backwards compatible way to solve this, in which case we'd need to do a v1 where users know about the difference in semantics for disabling a repository specified in a deb822 source.

Apologies for the giant wall of text below.

---

Currently, a `DebianRepository` has a `repotype` (e.g. `deb` or `deb-src`), a `uri` (e.g. https://nz.archive.ubuntu.com /ubuntu/ or https://security.ubuntu.com/ubuntu) and a unique `release` (e.g. `noble` or `noble-updates`). These are used to create its identifier in the `RepositoryMapping` in the form `$repotype-$uri-$release`.

A deb822 source entry can specify multiple of each of those components of the identifier. This PR iterates over them to create multiple `DebianRepository` objects with unique identifiers for each deb822 source entry. 

@NucciTheBoss pointed out that this is problematic when it comes to `DebianRepository` operations that expect to be able to modify the underlying source definition in the file, for example disabling the source. **Note:** is this the only operation? It doesn't seem like (re)enabling is an option, which is a great reduction in complexity

With the one-line-style source definition, disabling a source is as simple as commenting the line. An entire deb822 source entry can be enabled or disabled by setting the value of an `Enabled` field, but if a deb822 source entry is split into multiple `DebianRepository`s, then it's not clear how best to handle this.

---

In response to @NucciTheBoss's comments, I was leaning towards parsing a source into a single `Debian822Repository` that inherited from `DebianRepository` (or implemented the same protocol) instead of into multiple `DebianRepository`s, but I didn't consider the fact that such an object could have multiple of each of `repotype`, `uri`, and `source`. Not only is this not compatible with a consumer expecting to write `repo.repotype`, it makes it unclear how to include such an item in the `RepositoryMapping`, since it's not clear how to construct its identifier.

I think this makes parsing into multiple `DebianRepository` objects necessary for backwards compatibility.

---

I'd thought that ideally users of this library wouldn't even need to know that multiple source formats exist (one-line-style and deb822), but the semantics of enable/disable make this tricky.

One option would be to make a `v1` of the library, which could allow a simpler implementation, since we can just expose the source format to users with the knowledge that there are some differences between them (disable semantics, potential inline gpg keys, ... others?).

Here's my pitch on how to try for backwards compatibility. I think the best option is actually backwards incompatible, and the options that do maintain backwards compatibility of the python api need to mess with the underlying files too much to be reasonable. I'm open to better ideas.

1. parse deb822 entries into multiple `DebianRepository`s (actually backwards compatible subclasses), which keep an internal reference to an object that knows about the underlying deb822 style definition
2. on attempting to disable a single repository:
    - a) if there are in fact multiple repositories with the same underlying definition, raise an error. Provide a method (or keyword argument) to explicitly disable all the repositories with the same underlying source.
        - This is the sanest option I've come up with, but it *will* break users who are now suddenly reading in `.sources` files due to an invisible change based on the ubuntu version they're running on, and expecting to be able to do the same operations. So maybe we should just do an actual v1
    - b) edit the underlying source definition to disable the bits that are uniquely surfaced in this repository (!) ... but this could require a lot of editing
        - say the deb822 specifies 2 repotypes (`1` and `2`) and 2 releases (`a` and `b`). Say I didn't want the combination `1a`. The solution in deb822 style would be to write two entries, the first specifying the `1b` combination, and the second specifying the repotype `2` with releases `a` and `b`
        - this is a crazy option and I'd rather do a v1
    - c) delete/comment the original deb822 source style definition, and write out the corresponding one-per-line entries into a new file (and make the other repositories with the same deb822 source aware of this when they try to disable/etc )
        - a v1 would still be preferable to messing up the existing source files
    - d) on reading in a `.sources` file, write out a corresponding `.list` file and read that instead. When checking for `.source` files to read, if there's a matching `.list` file, just use that
        - probably the `.source` file needs to be hidden from `apt`, so that's added messing with existing source files, and I think a v1 would still be preferable

NucciTheBoss:
Thank you for your in-depth replies to my comments @james-garner-canonical! I appreciate it! We already chatted a bit over Matrix, but I also wanted to reply to your replies more in-depth as well :laughing: 

---

Re. your comments on the regex, I agree that it isn't perfect - there's tradeoffs to both approaches - I do personally find with regexes it's easier to figure out what's going on by slapping it into a regex tool with some provided test input to see how parsing text works compared to having to mentally walk through a custom parser. [However, I do agree with you that they get ugly fast](https://ihateregex.io/expr/ipv6) and can become more trouble than they're worth after a certain point.

However, even though the deb822 style is rather flexible, I do think we should draw the line somewhere with how the charm library expects it to be formatted to make our lives easier. Ideally charms are headless systems, so folks shouldn't be logging in and manually modifying the `*.sources` or `*.list` files. Otherwise we're going to end up trying to boil the ocean and duplicate the work of `python-debian` or other deb822 parsing tools :sweat_smile: 

---

Re. backwards compatibility with the `v0`, I wouldn't mind a `v1` release that's similar to this `apt` library I made for `cleantest` that's structurally similar to the `dnf` charm library (which should probably be deprecated): https://github.com/NucciTheBoss/cleantest/blob/main/src/cleantest/utils/apt.py. It could allow us to move beyond some of the technical debt/complexity around the v0 version of the apt charm library.

I think trying to maintain an internal register/reference that tracks a bunch of one-line sources in order to recombine them into a deb822 `*.sources` introduces unnecessary complexity. I think it would be better to just have to separate classes - `DebianOneLineStyleRepository` and `Debian822StyleRepository` - that inherit the common methods they need from an ABC like `DebianRepositoryBase`. That way they can be handled appropriately without it being necessary to force one style into another. Then the onus is put on the charm author to ensure that they're properly handling both source formats, or we only support the deb822 format and tell folks how to convert their one-line format to deb822 since it's easier to map one-line into deb822 than vice versa.

I my opinion it's fair to say _"hey if you want support for third-party repositories on Noble, you need to use v1 of this charm library because v0 doesn't support deb822 style sources"_ if it simplifies the implementation and reduces the maintenance burden of supporting newer Ubuntu releases that use the deb822 style source format rather than the legacy one-line style source format.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1315,47 +1409,293 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
-          default_filename: an (Optional) filename if the default is not desirable
+        """Add a new repository to the system. May have destructive or surprising behaviour.
+
+        Will write in the deb822 format if:
+            1. repo.filename has the '.sources' extension
+            2. repo.filename is unset (""), but /etc/apt/sources.list.d/ubuntu.sources exists
+        Otherwise the one-per-line style will be used.
+
+        WARNING: the default_filename keyword argument is provided for backwards compatibility
+        only. It is not used, and was not used in the previous revision of this library.
```
james-garner-canonical, line 1420: Can we just remove it? This method is called in one charm, and the default_filename argument isn't used there.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1315,47 +1409,293 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+
+        WARNING: in the one-per-line format case, will mutate repo.options to add the 'signed-by'
+        key if both options and gpg_file are truthy (for example if a gpg_key_filename and
+        non-empty options were provided at DebianRepository initialisation time). If options were
+        not provided or are empty, but a gpg_key is available, will silently fail to include it.
```
james-garner-canonical, line 1425: This is a bug. Potentially easy to fix. Doesn't affect the one user of this method, as the repository they pass doesn't use either `gpg_key` or `options`.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1315,47 +1409,293 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+
+        WARNING: if repo.filename is falsey, the new filename is calculated only from the repo's
+        uri and release, but repos are assumed to be uniquely identified by the combination of
+        repotype, uri, and release. Adding two repos with differing repotypes but identical uris
+        and releases will result in the second repo clobbering the file written by the first.
+        In this case, repo.filename must be set appropriately to avoid data loss.
```
james-garner-canonical, line 1431: This could be improved by including the repotype in the filename. Any concerns with compatibility when making this change? The lack of users points towards it being safe enough.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1315,47 +1409,293 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+
+        WARNING: if repo.filename is truthy, and that file exists, this method will clobber that
+        file with a single entry for the repo being added. Set it to a falsey value to have a new
+        filename derived from its uri and release (which will also clobber any existing file), or
+        set the filename that you want to write to -- or construct a new DebianRepository object
+        with the filename you want to write to.
+        For example: repo.filename = my_filename
+        For example: DebianRepository(uri=repo.uri, filename=my_filename, ...)
```
james-garner-canonical, line 1439: I don't love this behaviour. It seems especially bad that any `DebianRepository` `repo` created via a `RepositoryMapping` `mapping` will clobber the entire file it was loaded from with a single entry if you call `mapping.add(repo)`.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1315,47 +1409,293 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+
+        WARNING: if repo.enabled is falsey, the repository will be added in the disabled state.
+        Note that repo.disable() does not affect this value. If repo.enabled does not match the
+        value you expect, construct a new DebianRepository object with the appropriate value.
+        For example: DebianRepository(uri=repo.uri, enabled=my_value, ...)
```
james-garner-canonical, line 1444: This is probably ok, but a little weird, and potentially annoying since there's no api to enable a repository.

tests/unit/test_repo.py
```diff
@@ -146,3 +151,292 @@ def test_can_add_repositories_from_string_with_options(self):
...
+            assert list_repo.enabled == sources_repo.enabled
+            assert list_repo.repotype == sources_repo.repotype
+            assert list_repo.uri == sources_repo.uri
+            assert list_repo.release == sources_repo.release
+            assert list_repo.groups == sources_repo.groups
+            assert list_repo.gpg_key == sources_repo.gpg_key
+            assert (
+                list_repo.options  # pyright: ignore[reportUnknownMemberType]
+                == sources_repo.options  # pyright: ignore[reportUnknownMemberType]
+            )
```
james-garner-canonical, line 442: Unit test coverage of parsing deb822 is probably sufficient imo, but I'd like to add unit tests (or cases) for other operations touched by this PR (add, disable, gpg_key), as well as integration tests.

tests/unit/test_repo.py
```diff
@@ -146,3 +151,292 @@ def test_can_add_repositories_from_string_with_options(self):
...
+deb [signed-by=/usr/share/keyrings/ubuntu-archive-keyring.gpg] http://nz.archive.ubuntu.com/ubuntu/ noble-updates main restricted universe multiverse
+deb [signed-by=/usr/share/keyrings/ubuntu-archive-keyring.gpg] http://nz.archive.ubuntu.com/ubuntu/ noble-backports main restricted universe multiverse
+deb [signed-by=/usr/share/keyrings/ubuntu-archive-keyring.gpg] http://security.ubuntu.com/ubuntu noble-security main restricted universe multiverse
+"""
+
+
+class TestRepositoryMappingDeb822Behaviour(unittest.TestCase):
+    def test_iter_deb822_paragraphs_ubuntu_sources(self):
+        lines = ubuntu_sources_deb822.strip().split("\n")
+        paras = list(apt._iter_deb822_stanzas(lines))
```
benhoyt, line 202: We're testing a lot of internal methods here. Can we take more of an approach of testing these things *via* the public interface so that the tests aren't so dependent on implementation details?

tests/unit/test_repo.py
```diff
@@ -146,3 +151,292 @@ def test_can_add_repositories_from_string_with_options(self):
...
+        def iglob_nothing(s: str) -> typing.Iterable[str]:
+            return []
+
+        with patch("os.path.isfile", new=isnt_file):
+            with patch("glob.iglob", new=iglob_nothing):
+                repository_mapping = apt.RepositoryMapping()
+        assert not repository_mapping._repository_map
+
+        with patch(
+            "builtins.open", new_callable=mock_open, read_data=ubuntu_sources_deb822_with_comments
```
benhoyt, line 382: Similar here -- instead of patching `open`, can we just write a local file, which is fast and reliable?

tests/unit/test_repo.py
benhoyt: Yeah, a couple of integration tests would be good.

tests/unit/test_repo.py
NucciTheBoss: If you're trying to avoid destructive tests - e.g. non-isolated unit tests that will modify your host in some way - [`pyfakefs`](https://pypi.org/project/pyfakefs/) is quite good for mocking common filesystem operations such as read and writing files.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -837,7 +837,17 @@ def remove_package(
...
+    try:
+        subprocess.run(cmd, capture_output=True, check=True)
+    except CalledProcessError as e:
+        logger.error(
+            "%s:\nstdout:\n%s\nstderr:\n%s",
+            " ".join(cmd),
+            e.stdout.decode(),
+            e.stderr.decode(),
+        )
+        raise
```
james-garner-canonical, line 850: Just adding some useful logging

tests/integration/test_apt.py
```diff
@@ -44,53 +44,147 @@ def test_install_package_error():
...
+    assert repo_id not in apt.RepositoryMapping()
+    apt.update()
+    apt.remove_package("mongodb-org")
+    assert get_command_path("mongod")  # mongodb-org is a metapackage
+    subprocess.run(["apt", "autoremove"], check=True)
+    assert not get_command_path("mongod")
+
+
+def test_install_higher_version_package_from_external_repository():
+    repo_id = "deb-https://ppa.launchpadcontent.net/inkscape.dev/stable/ubuntu/-jammy"
```
james-garner-canonical, line 94: This test is kinda slow. Maybe better to pull the version out of the package listing without installing it?

tests/unit/test_repo.py
```diff
@@ -88,6 +89,7 @@ def test_can_disable_repositories(self):
             open(other.filename).readlines(),
         )
 
+    @pytest.mark.skip("RepositoryMapping.add now calls apt-add-repository")
```
james-garner-canonical, line 92: Probably these tests should be removed completely, or have their case moved into integration tests, rather than skipping

james-garner-canonical (commented):
My gut feeling is that this PR is too big for what it does

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1314,48 +1422,309 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+
+        WARNING: the default_filename keyword argument is provided for backwards compatibility
+        only. It is not used, and was not used in the previous revision of this library.
         """
-        new_filename = "{}-{}.list".format(
-            DebianRepository.prefix_from_uri(repo.uri), repo.release.replace("/", "-")
+        if not repo.enabled:
+            raise ValueError("{repo}.enabled is {value}".format(repo=repo, value=repo.enabled))
+        _add_repository(repo)
+        self._repository_map[_repo_to_identifier(repo)] = repo
```
james-garner-canonical, line 1448: Raising a `ValueError` here is technically a breaking change.

Previously, adding a disabled repository would write it to disk in commented out form. I don't think we can do that since we call `add-apt-repository` now.

It's just not clear what effect a user should expect from adding a disabled repository, hence raising an error, but for compatibility maybe it would be better to `logger.warning` and return early instead.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1314,48 +1422,309 @@ def _parse(line: str, filename: str) -> DebianRepository:
...
+
+    def __init__(
+        self,
+        message: str = "",
+        *,
+        file: str,
+        line: Optional[int],
+        key: str,
+        value: str,
+    ) -> None:
```
james-garner-canonical, line 1564: These custom exceptions can probably go back to just being InvalidSourceErrors

lib/charms/operator_libs_linux/v0/apt.py
benhoyt: Likely doesn't matter much, but I'd probably go with the warning+early return too.

tests/integration/test_apt.py
benhoyt: I wouldn't be too worried if the integration tests are only 5 minutes. Or maybe we can find a lighter package to install? If you can't find one fairly quickly, don't worry about it.

tests/unit/test_repo.py
benhoyt: Yeah, let's remove them.

lib/charms/operator_libs_linux/v0/apt.py
benhoyt: Seems fine to leave as is.

benhoyt (approved):

NucciTheBoss:
@james-garner-canonical @benhoyt can confirm that these changes work on our end for enabling Noble support in our HPC charms. Thanks for working on this 🤩

james-garner-canonical:
> @james-garner-canonical @benhoyt can confirm that these changes work on our end for enabling Noble support in our HPC charms. Thanks for working on this 🤩

That's great to hear, @NucciTheBoss! Merged now, thanks for your help with this

PR #138, opened 2024-11-25
Use snapd REST api to change snap config

weiiwang01:
Use the snapd REST API to change snap configurations instead of using the `snap` command.

This improves security in case a secret in command line arguments are logged.

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -343,7 +343,7 @@ class Snap(object):
         """
         if not typed:
             config = {k: str(v) for k, v in config.items()}
-        self._snap_client.put_snap_config(self._name, config)
+        self._snap_client._put_snap_conf(self._name, config)
 
     def unset(self, key) -> str:
         """Unset a snap configuration value.
@@ -774,12 +774,15 @@ class SnapClient:
             return self._wait(response["change"])
         return response["result"]
 
-    def _wait(self, change_id: int) -> JSONType:
+    def _wait(self, change_id: str, timeout=300) -> JSONType:
         """Wait for an async change to complete.
 
         The poll time is 100 milliseconds, the same as in snap clients.
         """
+        deadline = time.time() + timeout
         while True:
+            if time.time() > deadline:
+                raise TimeoutError(f"timeout waiting for snap change {change_id}")
             response = self._request("GET", f"changes/{change_id}")
             status = response["status"]
             if status == "Done":
@@ -840,7 +843,7 @@ class SnapClient:
         """Query the snap server for apps belonging to a named, currently installed snap."""
         return self._request("GET", "apps", {"names": name, "select": "service"})
 
-    def put_snap_config(self, name: str, conf: Dict[str, Any]):
+    def _put_snap_conf(self, name: str, conf: Dict[str, Any]):
         """Set the configuration details for an installed snap."""
         return self._request("PUT", f"snaps/{name}/conf", body=conf)
 
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -64,6 +64,7 @@ import re
 import socket
 import subprocess
 import sys
+import time
 import urllib.error
 import urllib.parse
 import urllib.request
@@ -83,7 +84,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 7
+LIBPATCH = 8
 
 
 # Regex to locate 7-bit C1 ANSI sequences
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -332,19 +333,17 @@ def get(self, key: Optional[str], *, typed: bool = False) -> Any:
 
         return self._snap("get", [key]).strip()
 
-    def set(self, config: Dict[str, Any], *, typed: bool = False) -> str:
+    def set(self, config: Dict[str, Any], *, typed: bool = False) -> None:
```
benhoyt, line 336: Technically this is a breaking change, but it previously returned whatever progress garbage that `snap set` printed to stdout, so charmers almost certainly weren't using it. So we're okay with changing this type to None. @james-garner-canonical to double check charm usage.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -818,6 +840,10 @@ def get_installed_snap_apps(self, name: str) -> List:
         """Query the snap server for apps belonging to a named, currently installed snap."""
         return self._request("GET", "apps", {"names": name, "select": "service"})
 
+    def put_snap_config(self, name: str, conf: Dict[str, Any]):
```
benhoyt, line 843: Let's make this private `_put_snap_config`.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -770,7 +769,30 @@ def _request(
             headers["Content-Type"] = "application/json"
 
         response = self._request_raw(method, path, query, headers, data)
-        return json.loads(response.read().decode())["result"]
+        response = json.loads(response.read().decode())
+        if response["type"] == "async":
+            return self._wait(response["change"])
+        return response["result"]
+
+    def _wait(self, change_id: int) -> JSONType:
```
benhoyt, line 777: Let's have a maximum timeout here, maybe 300s (5 minutes).

tests/integration/test_snap.py
```diff
@@ -65,7 +66,11 @@ def test_snap_refresh():
 def test_snap_set_and_get_with_typed():
     cache = snap.SnapCache()
     lxd = cache["lxd"]
-    lxd.ensure(snap.SnapState.Latest, channel="latest")
+    try:
```
benhoyt, line 69: From Weii: this is to address instability in the GitHub Runners when running these tests.

tests/unit/test_snap.py
```diff
@@ -902,28 +902,24 @@ def fake_snap(command: str, optargs: Optional[Iterable[str]] = None) -> str:
         with self.assertRaises(TypeError):
             foo.get(None)  # pyright: ignore[reportArgumentType]
 
-    @patch("charms.operator_libs_linux.v2.snap.subprocess.check_output")
-    def test_snap_set_typed(self, mock_subprocess):
+    @patch("charms.operator_libs_linux.v2.snap.SnapClient.put_snap_config")
```
benhoyt, line 905: Let's also add a test or two that tests the logic in the new `_wait` method.

lib/charms/operator_libs_linux/v2/snap.py
benhoyt: Also, it looks like `response["change"]` is actually a string (it only happens to be an integer inside the string). See [here](https://github.com/canonical/snapd/blob/4dd5dedffe85b9de33ca537a482b1787601ec262/daemon/api_snap_conf.go#L172) and [here](https://github.com/canonical/snapd/blob/4dd5dedffe85b9de33ca537a482b1787601ec262/overlord/state/change.go#L253).

benhoyt (changes requested):

lib/charms/operator_libs_linux/v2/snap.py
weiiwang01: Changed to private, thanks!

lib/charms/operator_libs_linux/v2/snap.py
weiiwang01: Renamed, thanks!

lib/charms/operator_libs_linux/v2/snap.py
weiiwang01: Added timeout and changed the typing for `change_id`, thanks!

tests/unit/test_snap.py
weiiwang01: Added, thanks!

benhoyt (approved):
Looks good to me, thanks @weiiwang01.

I haven't gone over the new tests with a fine-toothed comb. Added @james-garner-canonical as a second reviewer.

tests/unit/test_snap.py
```diff
@@ -697,6 +700,141 @@ def test_request_raw_bad_response_raises_snapapierror(self):
...
+                                "ready-time": "2024-11-28T20:06:41.797440022+00:00",
+                            },
+                        }
+                    ).encode("utf-8")
+                )
+            raise RuntimeError("unknown request")
+
+        client = snap.SnapClient()
+        with patch.object(client, "_request_raw", _request_raw), patch.object(time, "sleep"):
+            client._put_snap_conf("test", {"foo": "bar"})
```
james-garner-canonical, line 791: I guess ideally we'd call the public `Snap.set` instead of the private `SnapClient._put_snap_client` method, but that's not a blocker

james-garner-canonical (approved):
Look good, thanks Weii, I'll merge this in

Issue #139, opened 2024-12-12
[bug] Snap client `_wait` handles valid case as an error


Gu1nness:
## Description

The `SnapClient._wait` method ([here](https://github.com/canonical/operator-libs-linux/blob/f1952df48509dcdd51753eaa4a8ac7bb0357b27c/lib/charms/operator_libs_linux/v2/snap.py#L777)) considers a valid status `Do` as an error while it should not.

The [snapd REST API](https://snapcraft.io/docs/snapd-rest-api#heading--changes) specifies that the status can be `Do` for "Task is ready to start".
The filtering in the `SnapClient._wait` only considers the cases `Done` and `Doing`, `Valid` as valid cases and otherwise will raise a SnapError.
Similar to [this line](https://github.com/canonical/operator-libs-linux/blob/f1952df48509dcdd51753eaa4a8ac7bb0357b27c/lib/charms/operator_libs_linux/v2/snap.py#L790)), the status `Do` should sleep and wait for the task to start.

I can provide a patch if you want.

benhoyt:
Makes sense to me. A patch would be great, thanks! Out of interest, how did you run across this? If it was due to an error, do you remember what command/operation you were running at the time?

benhoyt:
Ah, I see you've opened #140 -- we'll take a look, thanks.

Gu1nness:
> Makes sense to me. A patch would be great, thanks! Out of interest, how did you run across this? If it was due to an error, do you remember what command/operation you were running at the time?

I'm working on the MongoDB charms and I'm refactoring everything.
As part of this refactoring, I'm using the `Snap.set` method to configure my snaps and I noticed that it started happening quite often (which never happened in the past), with the error message showing a status `Do`
A quick investigation on the REST API and the operator-libs led me to this :)

PR #140, opened 2024-12-12
fix: Task status can be `Do` and should not raise an error.

Gu1nness:
## Description

This PR fixes the issue #139 .
The task status can be `Do` and should not be considered an error ([see here](https://snapcraft.io/docs/snapd-rest-api#heading--changes) in the subsection 'Task Fields').

This status means that the task is ready to start but hasn't started yet.

## How is it tested?

I have added a case in the unit testing of the SnapClient.

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -84,7 +84,7 @@ LIBAPI = 2
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 8
+LIBPATCH = 9
 
 
 # Regex to locate 7-bit C1 ANSI sequences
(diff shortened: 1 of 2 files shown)


benhoyt (approved):
This looks good to me, thanks. For reference, here's the [snapd definitions](https://github.com/canonical/snapd/blob/5dfc211716455bfd9ed3cf1a08904bfa613ce316/overlord/state/change.go#L47-L51). I'd like @james-garner-canonical's review too just to keep him in the loop.

james-garner-canonical:
Looks great, thank you @Gu1nness ! I believe we'll need to bump the `LIBPATCH` number so that this change will get released on charmhub when we merge this PR

Gu1nness:
> Looks great, thank you @Gu1nness ! I believe we'll need to bump the `LIBPATCH` number so that this change will get released on charmhub when we merge this PR

My bad!
I completely forgot to bump the libpatch indeed, this is now done!

james-garner-canonical (approved):
Thanks for taking care of this!

PR #143, opened 2025-01-16
ci: update linting (following ops)

james-garner-canonical:
This PR updates `tox.ini` to use `ruff --preview` for all linting and formatting (dropping `black`), and have `codespell` run correctly, updating `pyproject.toml` accordingly.

The core of the PR is the changes to `tox.ini` and `pyproject.toml`. The linting rules for `ruff` have been updated to include all the rules that `ops` uses. To allow linting to continue to pass, any rules that are currently broken in a single file are ignored for that file, while rules that are broken in multiple files are ignored globally. These ignores are marked with a `FIXME` to distinguish them from intentional ignores that are copied from `ops`.

Changes to other files are either from `ruff format` or from `codespell`. There are typo corrections and a couple of small formatting changes to libraries, which are not significant enough to warrant bumping `LIB_PATCH` and triggering a release. There are also some bigger formatting changes for large dictionaries in tests.

Subsequent PRs will fix the ignored linting issues in individual libraries.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,33 +10,191 @@ show_missing = true
 minversion = "6.0"
 log_cli_level = "INFO"
 
-# Formatting tools configuration
-[tool.black]
-line-length = 99
-target-version = ["py38"]
-
 # Linting tools configuration
 [tool.ruff]
 line-length = 99
+target-version = "py38"
 extend-exclude = ["__pycache__", "*.egg_info", "htmlcov"]
 
+# Ruff formatter configuration
+[tool.ruff.format]
+quote-style = "preserve"  # FIXME: use double for consistency with current codebase convention
+
 [tool.ruff.lint]
-select = ["E", "W", "F", "C", "N", "D", "I001"]
-extend-ignore = [
-    "D203",
-    "D204",
-    "D213",
-    "D215",
-    "D400",
-    "D404",
-    "D406",
-    "D407",
-    "D408",
-    "D409",
-    "D413",
+select = [
+    # flake8-builtins
+    "A",
+    # flake8-bugbear
+    "B",
+    # pyflakes-docstrings
+    "D",
+    # conventional names
+    "C",
+    # flake8-copyright
+    # FIXME: enable when we switch to --preview for format
+    # "CPY",
+    # Pycodestyle (error)
+    "E",
+    # Pyflakes
+    "F",
+    # isort
+    "I001",
+    # pep8-naming
+    "N",
+    # Perflint
+    "PERF",
+    # Ruff specific
+    "RUF",
+    # flake8-bandit
+    "S",
+    # flake8-simplify
+    "SIM",
+    # pyupgrade
+    "UP",
+    # Pycodestyle (warning)
+    "W",
+    # flake8-2020
+    "YTT",
 ]
-ignore = ["E501", "D107"]
-per-file-ignores = {"tests/*" = ["D100","D101","D102","D103","D104"]}
+ignore = [
+    # Use from err or from None explicitly
+    # FIXME
+    "B904",
+    # Missing docstring in magic method
+    "D105",
+    # Missing docstring in `__init__`
+    "D107",
+    # too many leading # before block comment
+    # FIXME
+    "E266",
+    # Line too long
+    # FIXME
+    "E501",
+    # try/except inside loop
+    # FIXME: lift or noqa individually as appropriate
+    "PERF203",
+    # Unused noqa
+    # FIXME
+    "RUF100",
+    # use unpacking instead of concatenation with lists
+    # FIXME
+    "RUF005",
+    # implicit Optional in typing
+    # FIXME: typing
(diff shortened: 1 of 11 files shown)


tox.ini
```diff
@@ -54,18 +52,24 @@ deps =
     -r{toxinidir}/requirements.txt
     pyfakefs==5.5.0
     dbus-fast==1.90.2
+allowlist_externals =
+    mkdir
+commands_pre =
+    mkdir --parents .report
```
tonyandrewmeyer, line 58: It seems odd to use `--parents` for a folder that's in the current folder. Is it ever possible that the parent would need to be created?

pyproject.toml
```diff
@@ -9,35 +10,209 @@ show_missing = true
...
+    ".ruff_cache",
+    ".vscode",
+    "build",
+    "htmlcov",
+    "venv",
+    "icon.svg",
+]
+ignore-words-list = [
+    "assertIn",
+    "fpr",
```
tonyandrewmeyer, line 216: Is this intended as a temporary ignore so that we don't have to edit lots of things in this PR? If so, could there be a comment indicating that?

If it's meant as a permanent ignore, I don't love that. I don't have much idea what `fpr` might mean. File pointer? False positive rate?

pyproject.toml
```diff
@@ -9,35 +10,209 @@ show_missing = true
...
+"tests/unit/test_sysctl.py" = [
+    # combine if branches using logical or
+    # FIXME
+    "SIM114",
+]
+
+[tool.ruff.lint.pydocstyle]
+convention = "google"
+
+[tool.ruff.lint.mccabe]
```
tonyandrewmeyer, line 198: What about dropping mccabe instead? I typically find that it doesn't add much value, and that complexity is better part of the human judgement in a review. I think @benhoyt is also not a fan.

pyproject.toml
```diff
@@ -9,35 +10,209 @@ show_missing = true
...
+
+[tool.ruff.lint.pyupgrade]
+# Preserve types, even if a file imports `from __future__ import annotations`.
+# Since we target py 3.8, we want to keep using List[str] and Optional[str]
+keep-runtime-typing = true
+
+[tool.ruff.lint.per-file-ignores]
+"lib/charms/operator_libs_linux/v0/apt.py" = [
+    # Use a context manager for opening files
+    # FIXME: false positive, noqa it
```
tonyandrewmeyer, line 122: Does this happen 3+ times in that file? If not then I think a noqa directive on the line is nicer than for the file. Same for the other file-specific ones below.

*Unless* you plan on a quick follow-up that fixes these, in which case I don't think it really matters.

tonyandrewmeyer (approved):
Looks good, thanks!

I think if the intention is that we're marking this as work to be done in the future but have no specific plans on when that would be, then it might be nicer to have a group like:

```
# FIXME: all of these should be fixed when time allows
r1
r2
r3 # Some specific note
```

Rather than a lot of lines that are only "# FIXME".

*However*, if the intention is that all/most of this is cleaned up say this cycle, then I think it's fine like it is.

tox.ini
james-garner-canonical: `--parents` is just to permit the folder to already exist -- it won't already exist in CI, but it probably will if you're running `tox -e unit` locally. I'll add a note

pyproject.toml
james-garner-canonical: There's actually just one usage in the `apt` lib. Afaik codespell doesn't allow you to ignore specific uses, so it seems like adding it to the `ignore-words-list` is the only option. What do you think?
```
# from gnupg2 docs: fpr :: Fingerprint (fingerprint is in field 10)
return re.search(r"^fpr:{9}([0-9A-F]{40}):$", out, re.MULTILINE).group(1)
```

pyproject.toml
james-garner-canonical: Will do, thanks

pyproject.toml
james-garner-canonical: Agreed, I 100% prefer inline noqa too, this is just to reduce the library files touched by this PR -- and I do intend to fix as many individual libs as I can in follow ups

pyproject.toml
james-garner-canonical: I'll merge with a note added to the line for now, but do lmk if there's a better way

james-garner-canonical:
Thanks for the thorough review and helpful comments, @tonyandrewmeyer ! The intention is to fix the majority of the ignores in subsequent PRs, but I agree that it would be best to group all the global ignores that are intended to be removed separately from the ones that are intended to stay, so I've made that change

pyproject.toml
tonyandrewmeyer: Ah, this is something in a regular expression. It would be nice if codespell was clever enough to ignore all of those (or everything other than group names) because we really have no control over it.

Yeah, since we cannot rename the field it doesn't seem like there's any other option here.

PR #144, opened 2025-01-17
ci: add static type checking

james-garner-canonical:
This PR adds the `static` `tox` environment, and runs this static type checking in CI. It uses `pyright` to type check select libraries that have been updated to pass static type checking -- currently only `snap.py`. The `pyright` configuration in `pyproject.toml` is copied from `ops`.

A subsequent PR will update `snap.py` to pass all the other currently disabled linting checks. Future PRs may update other libraries and test suites to also pass curently disabled linting and type checking.


lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -65,14 +67,46 @@
...
 from enum import Enum
 from subprocess import CalledProcessError, CompletedProcess
-from typing import Any, Dict, Iterable, List, Optional, Union
+from typing import Dict, Mapping, Sequence, Union
+
+if typing.TYPE_CHECKING:
+    from typing import (
+        Callable,
+        Iterable,
+        List,
```
dimaqq, line 83: Given `from __future__ import annotations` at the top, use `list[x]` over `List[x]`

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -65,14 +67,46 @@
...
+        Literal,
+        NoReturn,
+        Optional,
```
dimaqq, line 86: And `x|None` over `Optional[x]`.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -220,39 +263,39 @@ class Snap(object):
...
-        name,
+        name: str,
         state: SnapState,
         channel: str,
         revision: str,
         confinement: str,
-        apps: Optional[List[Dict[str, str]]] = None,
-        cohort: Optional[str] = "",
+        apps: list[JSONObject] | None = None,
+        cohort: str | None = "",
```
dimaqq, line 272: Maybe for a separate PR...
What's the point of allowing None if the default value is an empty string and explicit None is coerced to an empty string too?

dimaqq (commented):
A partial review today.
I didn't expect that would be this much code 🙈

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Thanks, I use `List[x]` just once for a type alias defined inside this `TYPE_CHECKING` branch. I thought about using `'list[x]'` (string quoting) to avoid the import. Do you have a preference in this case?

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Thanks, a few of these slipped through, fixed now

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: I think that when this library was written, `Optional` was used a as a type annotation where the author wanted to indicate an optional (non-required) argument, rather than intending to allow `None` as a value.

I'd like to fix/change this, but for runtime backwards compatibility we need to handle `None` as an argument, so I've kept all the `str | None` types. And passing `cohort=None` looks kind of nice.

I did change the `bool | None` types to `bool`, since the runtime logic never handles the `None` case differently/explicitly (just as a falsey value when branching on the argument).

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: I've removed the use of `List` and `Dict` now, thanks for pointing this out

pyproject.toml
```diff
@@ -198,6 +198,21 @@ keep-runtime-typing = true
 [tool.ruff.lint.pydocstyle]
 convention = "google"
 
+[tool.pyright]
+include = ["lib/charms/operator_libs_linux/v*/*.py", "tests/*/*.py"]
+pythonVersion = "3.8"  # check no python > 3.8 features are used
+pythonPlatform = "All"
+typeCheckingMode = "strict"
```
benhoyt, line 205: I think we've kinda regretted using "strict" in Ops, as it can cause more churn than needed. Or I could be remembering wrong. I wonder how many of the `reportFoo = false`s below would go ahead if we used normal rather than strict mode?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -65,14 +67,56 @@
...
 import urllib.parse
 import urllib.request
-from collections.abc import Mapping
 from datetime import datetime, timedelta, timezone
 from enum import Enum
 from subprocess import CalledProcessError, CompletedProcess
-from typing import Any, Dict, Iterable, List, Optional, Union
+from typing import Mapping, Sequence, Union
+
+if typing.TYPE_CHECKING:
```
benhoyt, line 79: Can you remind me why we do this inside an `if typing.TYPE_CHECKING` block again?

lib/charms/operator_libs_linux/v2/snap.py
benhoyt: I guess we could change the default value of the arg to `None`?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -91,38 +135,47 @@
...
-JSONType = Union[Dict[str, Any], List[Any], str, int, float]
+# we use a non-null json type
+# because setting snap config values to null removes the key
+# so a null value can never be returned
+JSONObject = Mapping[str, 'JSONType']
+JSONType = Union['JSONObject', Sequence['JSONType'], str, int, float, bool]
+# we also need a nullable json type
+# because users can set snap config values to null
+NullableJSONObject = Mapping[str, 'NullableJSONType']
+NullableJSONType = Union[
```
benhoyt, line 155: I'm wondering whether these pay their way.

benhoyt (commented):
Looks reasonable at a quick glance. However, the actual changes are harder to see due to all the modernisations like `typing.List[T]` to `list[T]` and `Optional[T]` -> `T | None`. Could we do those changes in a separate PR first so we can see the "real" changes more easily?

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: It seemed cleaner to me to put all names that are only accessed in type annotations behind an `if typing.TYPE_CHECKING` because at runtime all those annotations will just be strings anyway. This avoids populating the library's namespace with names that aren't used at runtime.

One potential disadvantage is that if a library user want to refer to anything defined inside this block in their own type annotations, they won't be able to do `from library import name`, they'll have to do `import library; library.name`.

I'm happy to get rid of the `if typing.TYPE_CHECKING` though, maybe that would be more readable.

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Yeah it might be better to just have one `JSONType`, since users will typically need to `cast` or narrow the values of any returned `JSONType` anyway.

pyproject.toml
james-garner-canonical: Interesting. Looking [here](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#diagnostic-settings-defaults) it does seem like there are a lot of rules that we would lose if we dropped to "basic" type checking. It would indeed remove the need for the manually disabling the rules that we disable currently though.

I also [learned](https://github.com/microsoft/pyright/discussions/4715) that if we want to do gradual type annotations for this repository, it might be worth setting things to "basic" by default, and manually marking individual libraries (like snap) as being compliant with "strict" type checking.

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: I think that would be fine, as it's doesn't change the value that the attribute ends up being. I feel like you lose a little bit of information though as the default of `""` is what the value will end up being.

james-garner-canonical:
> Looks reasonable at a quick glance. However, the actual changes are harder to see due to all the modernisations like `typing.List[T]` to `list[T]` and `Optional[T]` -> `T | None`. Could we do those changes in a separate PR first so we can see the "real" changes more easily?

Sounds good @benhoyt, I'll do a separate PR tomorrow just modernising the type annotations without adding any new ones, and after we merge that in then we can pick up on this PR.

tox.ini
```diff
@@ -44,6 +44,17 @@ commands =
...
+[testenv:static]
+description = Run static type checker
+allowlist_externals = echo
+deps =
+    pyright==1.1.385
+commands =
+    echo 'NOTE: Only manually specified libs are being static type checked currently'
+    #pyright {[vars]lib_dir} {posargs}  # FIXME: enable static analysis of all libs
+    pyright lib/charms/operator_libs_linux/v2/snap.py {posargs}
+    #pyright {[vars]test_dir} {posargs}  # FIXME: enable static analysis of tests
```
dimaqq, line 56:
```suggestion
    echo 'NOTE: Only manually specified libs are being static type checked currently'
    pyright lib/charms/operator_libs_linux/v2/snap.py {posargs}
```

maybe?

pyproject.toml
```diff
@@ -198,6 +198,21 @@ keep-runtime-typing = true
...
+reportIncompatibleMethodOverride = false
+reportImportCycles = false
+reportMissingModuleSource = false
+reportPrivateUsage = false
+reportUnnecessaryIsInstance = false
+reportUnnecessaryComparison = false
+reportUnnecessaryTypeIgnoreComment = "error"
+disableBytesTypePromotions = false
```
dimaqq, line 213: I'm not sure what these settings do.
Do we need them?

pyproject.toml
```diff
@@ -198,6 +198,21 @@ keep-runtime-typing = true
...
+stubPath = ""
```
dimaqq, line 214: Do we need to stub out the stub path?

lib/charms/operator_libs_linux/v2/snap.py
dimaqq: Let's get rid of `if typing.TYPE_CHECKING` in this case.

I'd use it only cases like circular imports, that is importing real things for types only.

lib/charms/operator_libs_linux/v2/snap.py
dimaqq: There's no need for Union here, use `|`.

I'm not an expert, but ran into describing the JSON types in python-libjuju, and I believe we've settled on this:

```py
_JsonLeaf: TypeAlias = "None | bool | int | float | str"
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -158,32 +211,32 @@ class _Cache(object, metaclass=MetaCache):
...
 
     @property
-    def name(self):
+    def name(self) -> str:
         """Return a string representation of the model plus class."""
         return "<{}.{}>".format(type(self).__module__, type(self).__name__)
 
     @property
-    def message(self):
+    def message(self) -> str:
```
dimaqq, line 224: Without an explicit `__init__`, there's no way of knowing what type the arguments were and in fact if there were arguments at all.

Maybe an explicit `__init__` is actually the best solution?

Or is it the case that Error is never used directly?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -311,7 +364,15 @@ def _snap_daemons(
         except CalledProcessError as e:
             raise SnapError("Could not {} for snap [{}]: {}".format(args, self._name, e.stderr))
 
-    def get(self, key: Optional[str], *, typed: bool = False) -> Any:
+    @typing.overload
+    def get(self, key: None, *, typed: Literal[False] = False) -> NoReturn: ...
```
dimaqq, line 368: That's slightly out of touch with the implementation.

Implementation is `if not key: raise ...`, so both None and empty string would be marked NoReturn.

Is the implementation correct?

Arguably, the types can't be fixed with `Literal[""]` if the value comes from an expression, so maybe these overloads are the best possible?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -653,19 +711,22 @@ def confinement(self) -> str:
...
         self._update_snap_apps()
-        services = {}
+        services: dict[str, SnapServiceDict] = {}
         for app in self._apps:
             if "daemon" in app:
-                services[app["name"]] = SnapService(**app).as_dict()
+                name = typing.cast(str, app["name"])
+                snap_service = SnapService(**app)  # pyright: ignore[reportArgumentType]
+                service_dict = snap_service.as_dict()
+                services[name] = service_dict
```
dimaqq, line 729: I feel like this could be simplified.
Though, arguably not over-simplified.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -982,11 +1071,25 @@ def add(
     if isinstance(state, str):
         state = SnapState(state)
 
-    return _wrap_snap_operations(snap_names, state, channel, classic, devmode, cohort, revision)
+    return _wrap_snap_operations(
+        snap_names=snap_names,
+        state=state,
+        channel=channel if channel is not None else "",
+        classic=classic if classic is not None else False,
```
dimaqq, line 1078: It's `classic: bool = False` up top.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -982,11 +1071,25 @@ def add(
     if isinstance(state, str):
         state = SnapState(state)
 
-    return _wrap_snap_operations(snap_names, state, channel, classic, devmode, cohort, revision)
+    return _wrap_snap_operations(
+        snap_names=snap_names,
+        state=state,
+        channel=channel if channel is not None else "",
```
dimaqq, line 1077:
```suggestion
        channel=channel or "",
```

seems to be equivalent, wdyt?

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Ah I guess this works with `TypeAlias` from `typing_extensions`, excellent, thanks

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Makes sense. I was just working on doing this, when I realised that I'm also putting some `typing_extensions` imports behind `typing.TYPE_CHECKING` since the charm-lib can't depend on `typing_extensions` at runtime. This in turn means that runtime use of any symbols from `typing_extensions` also need to be in the `TYPE_CHECKING` block. In this case that's just `ParamSpec`. I think it still makes sense to move everything else out of the block though

pyproject.toml
james-garner-canonical: I just copied it from `ops` blindly haha. I guess it's to silence a warning in CI -- if `stubPath` isn't set, `pyright` will look for a directory named `typings` and emit a warning if it isn't found. See:  https://github.com/microsoft/pyright/issues/777

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Good point! `Error` isn't used directly. I think an explicit `__init__` is a good idea.

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Good catch, thanks. I believe that the implementation does intend to handle `None` and `""` identically. I'll add `Literal[""]` to the annotations, but you're right, the `overload`s won't quite work correctly if we have an empty string coming from somewhere typed as `str`. I suspect that this is the best we can do though

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: I do like the original one liner. Is this going too far?
```py
services[app["name"]] = SnapService(**app).as_dict()  # type: ignore
```
Probably the correct approach is actually to try and get `app` typed as an appropriate `TypedDict`, which I guess will require doing some exploration of the json values you can actually get here

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: This is just to allow `None` at runtime for backwards compatibility since it used to be `classic: Optional[bool] = False`. Could do `bool(classic)` or `classic or False` here instead of `if classic is not None` -- do you think that would be better?

pyproject.toml
james-garner-canonical: Copied from `ops` but some aren't applicable here. I've simplified and documented the settings, thanks

pyproject.toml
james-garner-canonical: I'm leaning away from the "basic" by default approach for this PR, as I'm not clear how to configure rules per level without having to use comments to configure it in each file -- e.g. make unnecessary type ignore an error in "strict" but not in "basic". PRs adding coverage for additional libs can always make this change

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Unfortunately I think this would be a breaking change, and some users do things like `raise snap.SnapError()` in their tests (as does this library) ... which would get a nice `IndexError` if they caught that error and tried to access its `message` attribute, but presumably they don't

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: On the other hand, it's a private attribute, so let's just make the argument default None

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: I tried the `TypedDict` approach so that I could just prefix the original one liner with
```py
app = typing.cast(_SnapServiceInputDict, app)
```
but I can't make it work.

So I'm leaning towards leaving it as I have it currently, since the casting and ignoring is minimal and selective (if a bit verbose).

---

Here's the `TypedDict` for reference
```py
_SnapServiceInputDict = TypedDict(
    "_SnapServiceInputDict",
    {
        "name": str,
        "daemon": str,
        "daemon_scope": str,
        "daemon-scope": str,
        "enabled": bool,
        "active": bool,
        "activators": list[str],
    },
)
```
the error is on unpacking `app` in the one liner:
Argument of type "object" cannot be assigned to parameter "kwargs" of type "str | None" in function "__init__"

Here's a minimal reproduction of the same error:
```py
_D = TypedDict("_D", {"a": int, "b": str})
def foo(d: _D):
    def bar(a: int, **kwargs: str): ...
    bar(**d)
```

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Having thought about it more, previously classic was just passed as is (even if it was `None`), so I think it's fine to not make the runtime change I made here (ie follow Dima's suggestion -- thanks)

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: How I've ultimately resolved this is:
- for private methods, drop the `Optional` typing and use `""` as the default
- for public methods, continue to type keyword `str` and `int` arguments as `Optional`, set the argument default to `None`, and ensure that it isn't `None` by the time it's passed to a private method internally

lib/charms/operator_libs_linux/v2/snap.py
dimaqq: Perhaps SnapError.__init__ can get a default value?
I guess I'm fine either way.

lib/charms/operator_libs_linux/v2/snap.py
dimaqq: Hmmm maybe we're supposed to use the new `: *tuple` or `: **dict` notation?
Or typing.Unpack (from extensions obv) in the interim?

The closest I could find is this:
https://github.com/microsoft/pyright/issues/9199

If that's not it, how about opening a discussion thread at pyright?

dimaqq: FWIW My preference would be to reduce this indented block to 2 lines.

dimaqq (approved):

dimaqq:
My 2c about 2 PRs: this PR, albeit a bit involved is about the right size.

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Good call on the default argument, done

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: Thanks for the extra motivation to pursue this and the clue to try `typing.Unpack`. Changing the `kwargs` annotation from `str | None` to an `Unpack`ed `TypedDict` with just `daemon-scope: NotRequired[str]` allows the two line solution to work

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -227,39 +287,39 @@ class Snap(object):
...
         self._channel = channel
         self._revision = revision
         self._confinement = confinement
-        self._cohort = cohort
+        self._cohort = cohort or ""
         self._apps = apps or []
         self._snap_client = SnapClient()
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
```
benhoyt, line 307: What's the difference between `other: object` and `other: Any`?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -823,35 +903,36 @@ def _request_raw(
...
                 body = {}
                 message = "{} - {}".format(type(e2).__name__, e2)
             raise SnapAPIError(body, code, status, message)
         except urllib.error.URLError as e:
-            raise SnapAPIError({}, 500, "Not found", e.reason)
+            raise SnapAPIError({}, 500, "Not found", str(e.reason))
         return response
 
-    def get_installed_snaps(self) -> dict:
+    def get_installed_snaps(self) -> list[dict[str, JSONType]]:
```
benhoyt, line 918: `dict` was just plain wrong before?

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -864,35 +945,35 @@ def __init__(self):
...
-        if snap is None:
-            # The snapd cache file may not have existed when _snap_map was
-            # populated.  This is normal.
-            try:
-                self._snap_map[snap_name] = self._load_info(snap_name)
-            except SnapAPIError:
-                raise SnapNotFoundError("Snap '{}' not found!".format(snap_name))
-
-        return self._snap_map[snap_name]
+        snap = self._snap_map.get(snap_name)
```
benhoyt, line 967: Thank you. Nice simplification here and below.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -985,9 +1088,23 @@ def add(
...
+        revision=revision or "",
+    )
 
 
+@typing.overload
+def remove(snap_names: str) -> Snap: ...
+# return a single Snap if snap name is given as a string
+@typing.overload
+def remove(snap_names: list[str]) -> Snap | list[Snap]: ...
+# may return a single Snap or a list depending if one or more snap names were given
```
benhoyt, line 1107: Ugh. I'm hereby convinced that `overload` points to a design issue. :-)

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: They're the same in that `__eq__` will accept any type for `other`, but differ in how `other` is typed inside the method.

`object` means we don't need anything more specific than the base `object` -- no special attributes, methods, etc, any Python object is fine. The type checker will complain if we try to do anything to `other` that you couldn't do to any object.

`Any` means something like don't worry about type checking `other` -- just assume that it has whatever attributes or methods we need. `other` will satisfy the requirements of any other type annotation without complaints -- `len(other)`, `int(other)`, etc are all fine.

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: According to the existing unit tests, yes. `get_installed_snaps` is mocked [here](https://github.com/canonical/operator-libs-linux/blob/51fa361cf2abbff203e536c99927aaa06d536b1e/tests/unit/test_snap.py#L282), returning [this list](https://github.com/canonical/operator-libs-linux/blob/51fa361cf2abbff203e536c99927aaa06d536b1e/tests/unit/test_snap.py#L107). I just ran the methods manually on my machine and verified that the tests are correct and the original annotation is incorrect

lib/charms/operator_libs_linux/v2/snap.py
benhoyt: Ah yes, that makes sense -- thanks.

benhoyt:
Oh, and we should probably bump up `LIBPATCH` in snap.py.

james-garner-canonical:
> Oh, and we should probably bump up `LIBPATCH` in snap.py.

Ah yes, @benhoyt, the reason I didn't do that in this PR is that I'm planning one more PR touching snap.py to hit non-type annotation linting issues, like switching to f-strings, etc. Ok to leave the bump till then?

benhoyt:
> Ok to leave the bump till then?

Ah yep, totally.

PR #145, opened 2025-01-21
style: use annotations from the future for the snap lib

james-garner-canonical:
This PR mechanically translates annotations using symbols like `Union`, `List`, and `Dict` to the newer type annotation syntax, and unquotes forward references, in the snap library. This should have no runtime effects aside from making all type annotations strings at runtime, and removing some imported `typing` symbols from the module namespace. Merging this PR will make it easier to  diff #144 with main.


benhoyt (approved):
This is very helpful, thanks.

PR #151, opened 2025-01-29
style: update type hints and enable type checking in ci for apt.py

james-garner-canonical:
Enable static analysis of `apt.py` in ci. `LIBPATCH` will be bumped after one more follow up PR handling `ruff` linting for this lib.

### DIFF
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -108,9 +108,10 @@ import logging
 import os
 import re
 import subprocess
+import typing
 from enum import Enum
 from subprocess import PIPE, CalledProcessError, check_output
-from typing import Any, Iterable, Iterator, Mapping
+from typing import Any, Iterable, Iterator, Literal, Mapping
 from urllib.parse import urlparse
 
 logger = logging.getLogger(__name__)
@@ -197,7 +198,7 @@ class DebianPackage:
         self._state = state
         self._version = Version(version, epoch)
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
         """Equality for comparison.
 
         Args:
@@ -232,7 +233,7 @@ class DebianPackage:
     @staticmethod
     def _apt(
         command: str,
-        package_names: str | list,
+        package_names: str | list[str],
         optargs: list[str] | None = None,
     ) -> None:
         """Wrap package management commands for Debian/Ubuntu systems.
@@ -346,8 +347,10 @@ class DebianPackage:
     def _get_epoch_from_version(version: str) -> tuple[str, str]:
         """Pull the epoch, if any, out of a version string."""
         epoch_matcher = re.compile(r"^((?P<epoch>\d+):)?(?P<version>.*)")
-        matches = epoch_matcher.search(version).groupdict()
-        return matches.get("epoch", ""), matches.get("version")
+        result = epoch_matcher.search(version)
+        assert result is not None
+        matches = result.groupdict()
+        return matches.get("epoch", ""), matches["version"]
 
     @classmethod
     def from_system(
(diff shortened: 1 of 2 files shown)


lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -346,8 +347,10 @@ def fullversion(self) -> str:
     def _get_epoch_from_version(version: str) -> tuple[str, str]:
         """Pull the epoch, if any, out of a version string."""
         epoch_matcher = re.compile(r"^((?P<epoch>\d+):)?(?P<version>.*)")
-        matches = epoch_matcher.search(version).groupdict()
-        return matches.get("epoch", ""), matches.get("version")
+        result = epoch_matcher.search(version)
+        assert result is not None
+        matches = result.groupdict()
+        return matches.get("epoch", ""), matches["version"]
```
james-garner-canonical, line 353: This issue here is that `re.Pattern.search(...)` can return `None`. Presumably it never does on well formed version strings, so we can just `assert`.

Note also the change to returning `matches["version"]` instead of `matches.get("version")`, as the latter is also typed as possibly returning `None`, which contradicts the function signature. The [regex](https://regex101.com/r/biTRkD/1) match should always have an entry for `"version"` since it matches `.*`.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -422,32 +425,33 @@ def from_installed_package(
...
+                name=matches["package_name"],
+                version=split_version,
+                epoch=epoch,
+                arch=matches["arch"],
+                state=PackageState.Present,
+            )
+            if (pkg.arch == "all" or pkg.arch == arch) and (
+                version == "" or str(pkg.version) == version
+            ):
+                return pkg
```
james-garner-canonical, line 454: Originally the whole loop body is wrapped in a `try ... except AttributeError`, again due to `re.Pattern.search(...)` potentially returning `None` instead of a `re.Match`. This change explicitly checks if the `search` result is `None`, allowing the loop body to be dedented.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -496,11 +500,11 @@ def from_apt_cache(
...
-                vals["Package"],
-                split_version,
-                epoch,
-                vals["Architecture"],
-                PackageState.Available,
+                name=vals["Package"],
+                version=split_version,
+                epoch=epoch,
+                arch=vals["Architecture"],
+                state=PackageState.Available,
```
james-garner-canonical, line 507: Just using keyword arguments because why not

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -555,15 +559,15 @@ def _get_parts(self, version: str) -> tuple[str, str]:
         upstream, debian = version.rsplit("-", 1)
         return upstream, debian
 
-    def _listify(self, revision: str) -> list[str]:
-        """Split a revision string into a listself.
+    def _listify(self, revision: str) -> list[str | int]:
+        """Split a revision string into a list.
```
james-garner-canonical, line 563: Return type is a list that alternates between strings and ints. This will come up later

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -651,31 +657,38 @@ def _compare_revision_strings(self, first: str, second: str):  # noqa: C901
                 # explicitly raise IndexError if we've fallen off the edge of list2
                 if i >= len(second_list):
                     raise IndexError
+                other = second_list[i]
```
james-garner-canonical, line 660: Use a local variable to allow type narrowing -- and it's more efficient anyway

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -651,31 +657,38 @@ def _compare_revision_strings(self, first: str, second: str):  # noqa: C901
...
+                    if item > other:
                         return 1
-                    if item < second_list[i]:
+                    if item < other:
                         return -1
                 else:
                     # string comparison
-                    return self._dstringcmp(item, second_list[i])
+                    assert isinstance(other, str)
+                    return self._dstringcmp(item, other)
```
james-garner-canonical, line 674: This is that list of alternating integers and strings -- well a pair of them being compared

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -651,31 +657,38 @@ def _compare_revision_strings(self, first: str, second: str):  # noqa: C901
...
                 return 1
             return 1
         # rev1 is shorter than rev2 but otherwise equal, hence lesser
         # ...except for goddamn tildes
-        if second_list[len(first_list)][0][0] == "~":
+        # FIXME: bug?? we return -1 in both cases
+        # FIXME: first_list[len(second_list)] should be a string, why are we indexing to 0 twice?
+        if second_list[len(first_list)][0][0] == "~":  # type: ignore
             return -1
         return -1
```
james-garner-canonical, line 689: `type: ignore`  due to the `str | int` ambiguity -- and this is weird and potentially buggy so I'd like to save further changes for a separate PR (unit test coverage of these comparison functions is spotty)

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -698,36 +711,52 @@ def _compare_version(self, other) -> int:
...
+    def __lt__(self, other: Version) -> bool:
         """Less than magic method impl."""
         return self._compare_version(other) < 0
 
-    def __eq__(self, other) -> bool:
+    def __eq__(self, other: object) -> bool:
         """Equality magic method impl."""
+        if not isinstance(other, Version):
+            return False
         return self._compare_version(other) == 0
```
james-garner-canonical, line 722: Technically a behaviour change, so maybe this should be dropped -- maybe it's better to get an `AttributeError` when you try to compare a `Version` for equality with something else?

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -748,8 +777,6 @@ def add_package(
         update()
         cache_refreshed = True
 
-    packages = {"success": [], "retry": [], "failed": []}
```
james-garner-canonical, line 751: I separate these into individual variables for better typing

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -698,36 +712,52 @@ def _compare_version(self, other) -> int:
...
+@typing.overload
+def add_package(
+    package_names: str,
+    version: str | None = "",
+    arch: str | None = "",
+    update_cache: bool = False,
+) -> DebianPackage: ...
+@typing.overload
+def add_package(
+    package_names: str | list[str],
```
benhoyt, line 751: Isn't `str | list[str]` a superset of the first overload? How does it know which signature to pick in the case of `str`? Should this be just `list[str]` here?

lib/charms/operator_libs_linux/v0/apt.py
benhoyt: Simpler too!

lib/charms/operator_libs_linux/v0/apt.py
james-garner-canonical: You're 100% correct, it should be `list[str]`. I'm surprised `pyright` didn't complain -- I guess it's because the return type is also a superset

benhoyt (approved):
Thanks!

PR #152, opened 2025-01-30
style: remove linting ignores for apt.py

james-garner-canonical:
This PR removes the linting ignores for apt.py and fixes each resulting problem. Commit view may be helpful as one ignore is fixed in each commit, but the file view is pretty straightforward.


tonyandrewmeyer (approved):
Looks good to me - no notes :)

Thanks!

Issue #129, opened 2024-08-09
ensure does not properly support "classic" confinement.


dstathis:
When a charm is already installed, and you refresh to a classic revision using `ensure(revision=N, classic=True)`, the lib does not add `--classic` to the command and so it fails.

benhoyt:
Thanks for the report. Yes, the code definitely looks a bit buggy on that front. Do you have time to put up a fix PR? If not, we can take a look later this week.

PR #161, opened 2025-05-27
fix: refresh with classic

IronCore864:
Fix a bug where `ensure` with `classic=True` doesn't add the `--classic` flag to the command.

Closes https://github.com/canonical/operator-libs-linux/issues/129.

### DIFF
--- a/tests/unit/test_snap.py
+++ b/tests/unit/test_snap.py
@@ -387,6 +387,31 @@ class TestSnapCache(unittest.TestCase):
         )
         self.assertEqual(foo._cohort, "")
 
+    @patch('charms.operator_libs_linux.v2.snap.subprocess.check_output')
+    def test_refresh_classic(self, mock_subprocess: MagicMock):
+        """Test that ensure and _refresh succeed and call the correct snap commands."""
+        foo = snap.Snap(
+            name='foo',
+            state=snap.SnapState.Present,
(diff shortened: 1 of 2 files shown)


tests/unit/test_snap.py
```diff
@@ -387,6 +387,31 @@ def test_refresh_revision_devmode_cohort_args(self, mock_subprocess: MagicMock):
...
+        foo = snap.Snap(
+            name='foo',
+            state=snap.SnapState.Present,
+            channel='stable',
+            revision='1',
+            confinement='classic',
+            apps=None,
+            cohort='A',
+        )
+        foo.ensure(snap.SnapState.Latest, revision='2', classic=True)
```
james-garner-canonical, line 402: If `classic` is not set to `True` in the `ensure` call, is `snap refresh` still called with `--classic` due to `foo.confinement` already being `'classic'`?

james-garner-canonical:
Looks good, please bump the lib patch version to trigger a new library release on merge, and remove the coverage report from the PR (maybe add `.report` to `.gitignore`), and we can merge this.

james-garner-canonical (changes requested):
Looks good, a few comments, and please bump the lib patch version in `snap.py` to trigger a new library release on merge.

james-garner-canonical (approved):

PR #163, opened 2025-07-03
feat: instrument the apt charm library

dimaqq:
NOT READY YET

### DIFF
--- a/lib/charms/operator_libs_linux/v0/apt.py
+++ b/lib/charms/operator_libs_linux/v0/apt.py
@@ -114,7 +114,10 @@ from subprocess import PIPE, CalledProcessError, check_output
 from typing import Any, Iterable, Iterator, Literal, Mapping
 from urllib.parse import urlparse
 
+import opentelemetry.trace
+
 logger = logging.getLogger(__name__)
+tracer = opentelemetry.trace.get_tracer(__name__)
 
 # The unique Charmhub library identifier, never change it
 LIBID = "7c3dbc9c2ad44a47bd6fcb25caa270e5"
@@ -125,6 +128,8 @@ LIBAPI = 0
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
 LIBPATCH = 17
+
+PYDEPS = ["opentelemetry-api"]
 
 
 VALID_SOURCE_TYPES = ("deb", "deb-src")
@@ -249,7 +254,9 @@ class DebianPackage:
         try:
             env = os.environ.copy()
             env["DEBIAN_FRONTEND"] = "noninteractive"
-            subprocess.run(_cmd, capture_output=True, check=True, text=True, env=env)
+            with tracer.start_as_current_span("FIXME") as span:
+                span.set_attribute("argv", _cmd)
+                subprocess.run(_cmd, capture_output=True, check=True, text=True, env=env)
         except CalledProcessError as e:
             raise PackageError(
                 f"Could not {command} package(s) {package_names}: {e.stderr}"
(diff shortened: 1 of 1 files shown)


dimaqq:
cc @Perfect5th and @chanchiwai-ray who contributed to the library.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -126,6 +129,8 @@
 # to 0 if you are raising the major API version
 LIBPATCH = 17
 
+PYDEPS = ["opentelemetry-api"]
```
tonyandrewmeyer, line 132: I think it would be good to call this out in the doc as well. In the ideal case, charmers don't look at the source, and PYDEPS is not necessarily read by any tools, so it's easy to miss in other cases. (Although if they have recent ops, then they'll be fine, but we probably can't assume that yet).

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1107,12 +1118,14 @@ def disable(self) -> None:
                 " Please raise an issue if you require this feature."
             )
         searcher = f"{self.repotype} {self.make_options_string()}{self.uri} {self.release}"
-        with fileinput.input(self._filename, inplace=True) as lines:
-            for line in lines:
-                if re.match(rf"^{re.escape(searcher)}\s", line):
-                    print(f"# {line}", end="")
-                else:
-                    print(line, end="")
+        with tracer.start_as_current_span("disable source") as span:
```
tonyandrewmeyer, line 1121: What's the motivation for the span here? Is it just because it's interacting with external files? Or because you suspect this might take a significant amount of time? Something else?

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1191,8 +1206,10 @@ def _get_key_by_keyid(keyid: str) -> str:
             "https://keyserver.ubuntu.com" "/pks/lookup?op=get&options=mr&exact=on&search=0x{}"
         )
         curl_cmd = ["curl", keyserver_url.format(keyid)]
-        # use proxy server settings in order to retrieve the key
-        return check_output(curl_cmd).decode()
+        with tracer.start_as_current_span(curl_cmd[0]) as span:
+            span.set_attribute("argv", curl_cmd)
```
tonyandrewmeyer, line 1210: I'm not sure if we would want to have the key ID in the span attributes. It's not secret, but it does feel personal, so not a great fit.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -1289,11 +1309,12 @@ def __init__(self):
                 if not os.path.isfile(default_sources):
                     raise
 
-        # read sources.list.d
-        for file in glob.iglob(os.path.join(sources_dir, "*.list")):
-            self.load(file)
-        for file in glob.iglob(os.path.join(sources_dir, "*.sources")):
-            self.load_deb822(file)
+        with tracer.start_as_current_span("load sources"):
```
tonyandrewmeyer, line 1312: Same question as above - would you mind explaining why this gets a span? I'm not saying it shouldn't, just trying to better understand.

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -126,6 +129,8 @@
 # to 0 if you are raising the major API version
 LIBPATCH = 17
```
tonyandrewmeyer, line 130: This should be bumped - I think the practice is that every improvement gets a release published, because there are so few here.

tonyandrewmeyer (commented):
What about adding spans around the top-level public functions? (Not the ones in classes, just the functions at the module level.) It feels like that's the equivalent of putting in an `__init__` span in an Object style lib -- and that it would be useful to know when these things (adding a package, removing a package, etc) is happening - basically at the point(s) you're entering this lib.

I assume no new tests because it's awkward to do that when this doesn't use Scenario, and we're generally happy to trust that adding tracing spans like this is very low risk.

If I use this with an old version of ops, do the traces just silently stay in memory and then vanish after the hook? Do any warnings or anything show?

lib/charms/operator_libs_linux/v0/apt.py
dimaqq: My original thinking was that it has system-wide effect.

The secondary was that the file could be arbitrarily large and processing could take arbitrarily long. Then again, these files are never massive.

lib/charms/operator_libs_linux/v0/apt.py
dimaqq: You have a point.
Then again, this is a network request with a URL parameter.
It's pretty much bound to be logged on the server.

Note that the key id is the id of key being imported, that is some PPA being added (most likely, the library doesn't enforce that).
It's not a personal key.

dimaqq:
> What about adding spans around the top-level public functions? 

I've decided against that, because these functions lead directly to a subprocess call or file modification.

For a simple enough charm that only needs a single package from apt (hardcoded package name), it should be pretty obvious what happens. If some charm needs a couple of packages, it's still relatively easy to see what package was affected.

Only if some charm needs to handle many packages, or if there's another charm lib that wraps this (and package name is parametrised), would it really help.

I think that in those cases, that caller code should be instrumented instead.

Basically, I'm following the same pattern we've settled on in ops.

Do you buy this argument, @tonyandrewmeyer ?

tonyandrewmeyer (approved):

lib/charms/operator_libs_linux/v0/apt.py
```diff
@@ -98,6 +98,10 @@
     repo = DebianRepository.from_repo_line(line)
     repositories.add(repo)
 ```
+
+Dependencies:
+Note that this module requires `opentelemetry-api`, which your charm already has if
+the `ops` dependency is up to date.
```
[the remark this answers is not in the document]
dimaqq: That's a good point, I've rewritten this part.

james-garner-canonical (approved):

PR #165, opened 2025-08-11
feat: track snap version

skatsaounis:
Keep track of the snap version, if such version is available so that it could be used from a charm to set the workload version. 

For example, an operator charm could set the version like this:

```py
def _on_start(self, _event: ops.StartEvent) -> None:
    self.unit.status = ops.MaintenanceStatus("starting...")
    maas = SnapCache()["maas"]
    workload_version = maas.version if maas.version is not None else maas.revision
    self.unit.set_workload_version(workload_version)
```

### DIFF
--- a/lib/charms/operator_libs_linux/v2/snap.py
+++ b/lib/charms/operator_libs_linux/v2/snap.py
@@ -309,8 +309,8 @@ class Snap:
       - state: a `SnapState` representation of its install status
       - channel: "stable", "candidate", "beta", and "edge" are common
       - revision: a string representing the snap's revision
-      - version: a string representing the snap's version
       - confinement: "classic", "strict", or "devmode"
+      - version: a string representing the snap's version, if set by the snap author
     """
 
     def __init__(
@@ -319,19 +319,20 @@ class Snap:
         state: SnapState,
         channel: str,
         revision: str,
-        version: str,
         confinement: str,
         apps: list[dict[str, JSONType]] | None = None,
         cohort: str | None = None,
+        *,
+        version: str | None = None,
     ) -> None:
         self._name = name
         self._state = state
         self._channel = channel
         self._revision = revision
-        self._version = version
         self._confinement = confinement
         self._cohort = cohort or ""
         self._apps = apps or []
+        self._version = version
         self._snap_client = SnapClient()
 
     def __eq__(self, other: object) -> bool:
@@ -754,11 +755,6 @@ class Snap:
         return self._revision
 
     @property
-    def version(self) -> str:
-        """Returns the version for a snap."""
-        return self._version
-
-    @property
     def channel(self) -> str:
         """Returns the channel for a snap."""
         return self._channel
@@ -791,6 +787,11 @@ class Snap:
         """Report whether the snap has a hold."""
         info = self._snap("info")
         return "hold:" in info
+
+    @property
+    def version(self) -> str | None:
+        """Returns the version for a snap."""
+        return self._version
 
 
 class _UnixSocketConnection(http.client.HTTPConnection):
@@ -1055,9 +1056,9 @@ class SnapCache(Mapping[str, Snap]):
                 state=SnapState.Latest,
                 channel=i["channel"],
                 revision=i["revision"],
-                version=i["version"],
                 confinement=i["confinement"],
                 apps=i.get("apps"),
+                version=i.get("version"),
             )
             self._snap_map[snap.name] = snap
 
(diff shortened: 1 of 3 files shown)


lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -317,6 +319,7 @@ def __init__(
         state: SnapState,
         channel: str,
         revision: str,
+        version: str,
         confinement: str,
         apps: list[dict[str, JSONType]] | None = None,
         cohort: str | None = None,
```
james-garner-canonical, line 325: Since this is a public class, this addition to the constructor needs to be backwards compatible. Unfortunately that does mean that it will have to be an optional argument, which in turn means it probably has to be `None` able ... but at least we can make it keyword only to minimise ordering issues if we add additional arguments in future.
```suggestion
        state: SnapState,
        channel: str,
        revision: str,
        confinement: str,
        apps: list[dict[str, JSONType]] | None = None,
        cohort: str | None = None,
        *,
        version: str | None = None,
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -749,6 +753,11 @@ def revision(self) -> str:
         """Returns the revision for a snap."""
         return self._revision
 
+    @property
+    def version(self) -> str:
```
james-garner-canonical, line 757:
```suggestion
    def version(self) -> str | None:
```

tests/unit/test_snap.py
```diff
@@ -326,20 +326,20 @@ def test_raises_error_if_snap_not_running(self, mock_exists):
         self.assertIn("snapd is not running", ctx.exception.message)
 
     def test_can_compare_snap_equality(self):
-        foo1 = snap.Snap("foo", snap.SnapState.Present, "stable", "1", "classic")
-        foo2 = snap.Snap("foo", snap.SnapState.Present, "stable", "1", "classic")
+        foo1 = snap.Snap("foo", snap.SnapState.Present, "stable", "1", "v42", "classic")
+        foo2 = snap.Snap("foo", snap.SnapState.Present, "stable", "1", "v42", "classic")
```
james-garner-canonical, line 330:
```suggestion
        foo1 = snap.Snap("foo", snap.SnapState.Present, "stable", "1", "classic", version="v42")
        foo2 = snap.Snap("foo", snap.SnapState.Present, "stable", "1", "classic")
```

james-garner-canonical: We don't include `version` in `Snap.__eq__` -- which I think is a fine call, as we already compare `revision`, which should be appropriately precise. Let's assert that here by not including the `version` for one of the snaps we're comparing. I'd also suggest adding a test case for inequality where the revision numbers are different but other metadata is the same (including version -- unrealistic? but worth asserting I think).

lib/charms/operator_libs_linux/v2/snap.py
james-garner-canonical: The call sites in the unit tests will need to be updated accordingly.

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -1046,6 +1055,7 @@ def _load_installed_snaps(self) -> None:
                 state=SnapState.Latest,
                 channel=i["channel"],
                 revision=i["revision"],
+                version=i["version"],
```
james-garner-canonical, line 1058: Can we be sure that `"version"` is provided for all snaps? (e.g. [here](https://askubuntu.com/questions/915160/how-do-i-get-the-version-number-of-a-snap-package) is discussion of a snap where this field wasn't exposed).
```suggestion
                version=i.get("version"),
```

lib/charms/operator_libs_linux/v2/snap.py
```diff
@@ -1065,6 +1075,7 @@ def _load_info(self, name: str) -> Snap:
             state=SnapState.Available,
             channel=info["channel"],
             revision=info["revision"],
+            version=info["version"],
```
james-garner-canonical, line 1078:
```suggestion
            version=info.get("version"),
```

james-garner-canonical (changes requested):
Hi there, thanks for your contribution. The changes seem broadly reasonable, but there are a couple of changes that would be necessary before merging.

Firstly, let's validate this change in integration tests -- just checking that the exposed version field contains a string as expected in one of the existing tests would be sufficient. 

Secondly, adding a new, required argument to the public `Snap` class is not a backwards compatible change, particularly when added in the middle of other positional arguments. See separate comment for suggested resolution.

Also, I wonder if `version` is always guaranteed to be populated, or if it's up to the developer of the snap -- let's be defensive here.

Oh and if you could update the PR description with a description of the changes and the motivation, that would be great, thanks!

james-garner-canonical:
PS: don't forget to sign the CLA.

james-garner-canonical (approved):
Looks great, thanks!

PR #166, opened 2025-08-19
feat: add support for passing "all" arg to snap.logs

PietroPasotti:
the `snap logs` command supports an `all` parameter for the `-n` flag.

```
# from the command's own --help:
[logs command options]
          --abs-time   Display absolute times (in RFC 3339 format). Otherwise, display relative times up to 60 days, then YYYY-MM-DD.
      -n=              Show only the given number of lines, or 'all'. (default: 10)
      -f               Wait for new lines and print them as they come in.

```
This PR adds support for that (in the form of a type hint correction, the implementation is unchanged).


james-garner-canonical (approved):
Looks good to me, please bump LIBPATCH if you want to have this fix released after merging.

PR #167, opened 2025-08-21
docs: for apt, clarify that `add_package` doesn't raise `PackageNotFoundError`

dwilding:
When testing the apt library, I noticed that `add_package` doesn't raise `PackageNotFoundError` if any packages aren't found. Instead, the function raises `PackageError`. That's because `add_package` relies on a function `_add` to add each package, and `_add` catches `PackageNotFoundError` and returns a boolean instead.

This PR updates the relevant docs to clarify when `PackageNotFoundError` is raised. Matches https://github.com/canonical/charmlibs/pull/124


james-garner-canonical (approved):
Thanks for taking care of this

PR #171, opened 2026-02-25
Make wrap_snap_operations return the actual error.

crypticC0der:
Currently wrap_snap_operations just returns minimal error messages including only the names of the snaps that failed and not why. This is not particularly useful so I've expanded this to return include the actual error as well.


crypticC0der:
> Is this an improvement you're hoping to see in charms as soon as possible, or would waiting for the rewrite work?

I'm raising this just as I'm making the change locally in https://github.com/canonical/microovn-operator/pull/55 and i thought I'd pass along the update if you found it useful. I am in no real rush for the change to take place but please let me know when the rewrite is done with the new charm lib and I'll try move our products over

benhoyt:
Closing as this repo is being archived. Ideally users can move to charmlibs.snap.

